PR #101, opened 2014-03-05
Dirty "ensure => 'reinstalled'" support for plugin upgrade

p0deje:
This is pretty dirty as it __always__ removes plugin and installs new version. I could avoid this, but I didn't found a way to check the version of installed plugins using ES plugin tool.

It also changes behavior so `ensure => 'blah-blah'` does not remove plugin, only `ensure => absent` does.

### DIFF
--- a/README.md
+++ b/README.md
@@ -426,6 +426,14 @@ class { 'elasticsearch':
 }
 ```
 
+### Upgrade/reinstall plugin
+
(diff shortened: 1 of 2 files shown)


p0deje:
@elasticsearch-release Oh, hai ghprb!

electrical:
Closing in favour of https://github.com/elastic/puppet-elasticsearch/pull/213

PR #154, opened 2014-07-25
It 7388 01

dtaylor-desk:


electrical:
Decision internally has been made not to support upstart as newer versions of debian and ubuntu are using systemd.
Thank you for the contribution though.

PR #179, opened 2014-09-01
Hiera support

electrical:


electrical:
I've enabled the hiera support for rspec-puppet but getting weird errors.

this is one of the errors:

```
  1) elasticsearch on redhat-5-x86_64 main class tests should compile into a catalogue without dependency cycles
     Failure/Error: it { should compile.with_all_deps }
       error during compilation: Parameter user failed on Exec[install_plugin_mobz/elasticsearch-head]: Only root can execute commands as other users at /home/richard/projects/puppet/modules/puppet-elasticsearch/spec/fixtures/modules/elasticsearch/manifests/plugin.pp:116
     # ./spec/classes/000_elasticsearch_init_spec.rb:42:in `block (5 levels) in <top (required)>'

```

Any idea's ?

philipwigg:
Ugh. Yes.

Now you're creating an instance, your catalogue contains the Exec class for installing plugins.

You're specifying which user the Exec should run as which is something only root can do. As you're not root, the manifest won't compile.

But it's only a unit test so it shouldn't care, right?

A couple of other people have noticed this, see:-

https://tickets.puppetlabs.com/browse/PUP-2076

https://github.com/rodjek/rspec-puppet/issues/175

So I think it's a bug - I'm not really sure what the solution though I'm afraid...

electrical:
I have removed the check for compiling and works now :-) stupid stuff :-1:

electrical:
Only issue i'm running into now is that existing tests fail because the hiera config adds stuff to the params stuff that i'm not expecting in the tests. Any idea's?

I thought of adding a fact for the hostname and use that in the hiera config to ensure its only enabled then.
Any idea's ?

philipwigg:
Yeah that makes sense.

You don't even need to have the custom fact be 'hostname', you can have something more descriptive along the lines of 'hiera_test_data' or whatever so it's more obvious what it's doing?

As long as you include it in the hiera.config hierarchy it will work just the same. You could then make the .yaml filename match or at least relate to the scenario or the data in the file? single_hiera_instance.yaml, multiple_hiera_instance.yaml, etc.

What do you think?

electrical:
Oh that's a good one, forgot i can set a custom fact in there.
Thanks for the tip!

electrical:
Got the main parts working to test out the instances and plugins parts.
Do you have any idea's on testing plugins_hiera_merge and instances_hiera_merge ?
All i can think of is adding an other custom fact that loads up an other hiera file and then use that?

electrical:
I've just added some hiera spec tests.
Could you have a quick look if it covers most of it?

Cheers.

philipwigg:
Looks good to me! I like the defaultinstance combined with scenario.

electrical:
Thanks. Now i have to think up something for the acceptance testing with beaker.
Especially the differences between foss and PE make it a bit harder to manage the hiera.yaml file in a decent way.

electrical:
Managed to add some simple acceptance testing as well.
Going to see if it wants to bake in CI decently before merging it.

bflad:
Extremely excited that we will not have to manage this outside of this module. Thanks so much for all the work here!

PR #197, opened 2014-11-06
Closes #196

### LINKED ISSUE #196: Template absent fix and accept content instead of only file
When working with the elasticsearch::template class, there is one bug that I found and one improvement that I would like to make.

For the bug, when specifying an existing template to be "absent", the template is not removed from the server or the template directory.  This would be a simple change to ensure that the Exec["delete_template_${name}"] gets called to delete it from the server and a different "file" resource to ensure that it is deleted from the template_import directory.

For the improvement, instead of passing in a file uri handle, I would like to pass in a json string which is the content of the file.  I am doing this as an alternative way to control the content of the template via hiera for our application developers.

amoswood:


electrical:
Thanks for the delay in merging this. finally got more time for the puppet modules.
Took an other look an it looks great. will merge as is and add some tests to it later.

Thanks!

PR #220, opened 2015-01-05
Convert $config to an empty hash

ajlanghorn:
When using this, the 'undef' value of the $config parameter posed an issue
because Puppet interprets values as strings unless otherwise specified, yet
the Puppet merge function in instance.pp was expecting a hash.

To get around this, committing an empty hash as the default value.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -166,6 +166,7 @@
 #
 # === Authors
 #
+# * Andrew Langhorn <mailto:andrew@ajlanghorn.com>
 # * Richard Pijnenburg <mailto:richard.pijnenburg@elasticsearch.com>
 #
 class elasticsearch(
@@ -188,7 +189,7 @@ class elasticsearch(
   $init_defaults         = undef,
   $init_defaults_file    = undef,
   $init_template         = undef,
-  $config                = undef,
+  $config                = {},
   $datadir               = $elasticsearch::params::datadir,
   $plugindir             = $elasticsearch::params::plugindir,
   $plugintool            = $elasticsearch::params::plugintool,


electrical:
Hi,

WIth which versions did you test this?
I've ran rspec tests with a undef $config setting and shows no errors.
This is with puppet 3.2 and up.

ajlanghorn:
@electrical Running it on 3.6.2.

electrical:
Very weird i'm not seeing this issue with the rspec tests.
I'll add a test on the acceptance testing and see if it shows up.

ajlanghorn:
Thanks @electrical!

electrical:
my local acceptance test is passing.
Can you tell me which version of puppetlabs-stdlib you are using?

electrical:
I've been trying out different options with puppet versions and puppetlabs-stdlib versions and have been unable to reproduce the issue.
could you check it further?

ajlanghorn:
Using Puppet 3.6.2 and `puppetlabs/stdlib 4.1.0` here. I'll see about testing against different versions of the Puppet Gem.

electrical:
Ran the acceptance test with 3.6.2 and stdlib 4.1.0 and unable to reproduce the issue.
can you check the ruby version? that might be the last difference i can think of.

ajlanghorn:
Running 1.9.3p484 here; I'll try with a newer version.

electrical:
it could be that ruby 1.9.3 is doing something weird. i did see that stdlib 4.2.0 has an extra check to see if its a string and empty, that might fix your issue with that ruby version ( just a guess at this point )

ajlanghorn:
@electrical I think you're right - using the latest and greatest `puppetlabs/stdlib` version seems to work fine. I'm going to close this PR off for the meantime; if I need to, I'll re-open it. Thanks for your help!

PR #224, opened 2015-01-08
Fix conflicting data.path in config and datadir option

### LINKED ISSUE #186: elasticsearch::instance doesn't handle conflicting parameters
The following code snippet doesn't fail, but randomly switches ES datadir from `/foo` to `/bar` and back :
```
elasticsearch::instance { 'auth-logs':
  datadir    => '/foo',
  config     => {
    'path'   => {
      'data' => '/bar',
    },
  },
}
```

I believe it should either fail or discard one of the 2.

electrical:
In some cases people would define a data.path setting in the config and use the datadir option.
We should detect this and force the datadir option to be leading.

Fixes #186

### DIFF
--- a/manifests/instance.pp
+++ b/manifests/instance.pp
@@ -168,7 +168,15 @@ define elasticsearch::instance(
       $main_config = { }
     }
 
-    $instance_datadir_config = { 'path.data' => $instance_datadir }
+    if(has_key($instance_config, 'path.data')) {
+      $instance_datadir_config = { 'path.data' => $instance_datadir }
+    } elsif(has_key($instance_config, 'path')) {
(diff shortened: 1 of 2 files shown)


erik-stephens:
I got tripped up by this.  I defined config['path']['data'] but not `datadir`.  It was quietly using the default datadir which I find unintuitive.  I recommend failing, or maybe just a warning, with a msg to define `datadir` [here](https://github.com/elasticsearch/puppet-elasticsearch/blob/c1c88ac68ee13319a588d3214460957b961ef6db/manifests/instance.pp#L171).  

Probably not worth addressing unless others have a similar experience.  Just a heads up & recommendation in case pops up again.

Issue #221, opened 2015-01-05
puppet newbie -- testing and not seeing config changes


SteveDevOps:
I'm finally getting communications happening and able to invoke a manifest specifying elasticsearch install. However my config param aren't showing up in the default .yml. What am I missing? -- I'm testing on centos 7 vms, running PE 3.7.

My custom manifest:

```
node 'xxxx' {

	group { 'esadmin':
		ensure => 'present'
	}

	user { 'esadmin':
		ensure => 'present',
		groups => 'esadmin'
	}

	class {'elasticsearch':
		java_install => true,
		manage_repo => true,
		repo_version => '1.4',
		datadir => '/var/data/elasticsearch',
		elasticsearch_user => 'esadmin',
		elasticsearch_group => 'esadmin',
		config => {
			'node.name' => $::fqdn,
			'cluster' => {
				'name' =>	'esearch',
				'routing.allocation.awareness.attributes' => 'rack'
			},
			'index' => {
				'number_of_replicas' => '1',
				'number_of_shards' => '4'
			},
			'network' => {
				'host' => $::ipaddress
			}
		}




	}



}

```

electrical:
Hi,

You will need to define an instance as specified in https://github.com/elasticsearch/puppet-elasticsearch#instances
Hope it helps.

SteveDevOps:
Thanks.. ok got that in place, and accepted by agent. Now I'm confused on how to access the "instance" via url. I have the default service off, I found the extra instance service on my centos vm,  and I'm re-specifying http.port as 9200 under my config => for instance.

electrical:
Hi,

You should be able to access it via the same IP and port as you would do with the default instance.

SteveDevOps:
electrical -- I get nothing.. If I switch back to using default elasticsearch service it works but obviously doesn't have my puppet-specified config. 

```
[root@pnode-test elasticsearch]# tree /etc/elasticsearch
/etc/elasticsearch
├── elasticsearch.yml
├── logging.yml
├── pnode-test
│   ├── elasticsearch.yml --- this one has my config for instance
│   └── logging.yml
└── templates_import
```

```
[root@pnode-test elasticsearch]# chkconfig --list

Note: This output shows SysV services only and does not include native
      systemd services. SysV configuration data might be overridden by native
      systemd configuration.

      If you want to list systemd services use 'systemctl list-unit-files'.
      To see services enabled on particular target use
      'systemctl list-dependencies [target]'.

[...]

Issue #206, opened 2014-12-12
/var/run/elasticsearch folder isn't chown'd


chriswhite199:
When configuring a different es user / group through the elasticsearch_user and elasticsearch_group params, the pid folder isn't chown'd to the new user:group, which prevents the instance from starting as a service.

### Config:

    class { 'elasticsearch':
      package_url => 'puppet:///modules/elasticsearch/elasticsearch-1.1.1.noarch.rpm',
      config => {
        'node' => {
          'name' => $::fqdn
        },
        'index'                => {
            'number_of_replicas' => '1',
            'number_of_shards'   => '5'
          },
          'network'              => {
            'host'               => $::ipaddress
          },
          'cluster'              => {
            'name'               => "esearch",
          }
      },
      elasticsearch_user => 'elastics',
      elasticsearch_group => 'elastics',
    }

    elasticsearch::instance { 'es-01': }

### Starting instance:

    Starting elasticsearch-es-01:                              [  OK  ]
    [root@2c9efb74cbb1 manifests]# {1.1.1}: pid Failed ...
    - FileNotFoundException[/var/run/elasticsearch/elasticsearch-es-01.pid (Permission denied)]

### Directory listing:

    [root@2c9efb74cbb1 puppet]# ll /var/run/elasticsearch -a
    total 8
    drwxr-xr-x. 2 elasticsearch elasticsearch 4096 Apr 16  2014 .
    drwxr-xr-x. 9 root          root          4096 Dec 12 03:44 ..

electrical:
Hi,

I've just merged a fix into the master branch which should solve the rights issues.

Cheers.

electrical:
Testing showed this issue to be fixed in the master branch.
Closing the issue for now, could you open a new one if its not fixed? Thanks!

PR #225, opened 2015-01-08
Fix rights on different directories and files

### LINKED ISSUE #201: Repository based install won't chown relevant directories.
I'm installing elasticsearch via repository, setting elasticsearch_user and elasticsearch_group and also ES_USER and ES_GROUP, all set to the same value.
However, the log-directory, the data-directory, the config directory, the var/run/elasticsearch directory all are owned by elasticsearch:elasticsearch. This has the effect that elasticsearch won't start.
My current workaround is chown-ing all these directories myself.

electrical:
We didn't manage all the files so when switching to an other user it could break
Also some plugins copy files to other directories of Elasticsearch and broke because of owner issues

Fixes #201 #206 #221

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -56,20 +56,17 @@ class elasticsearch::config {
 
     file { $elasticsearch::params::homedir:
       ensure  => 'directory',
-      mode    => '0644',
       recurse => true
     }
 
     file { $elasticsearch::plugindir:
       ensure  => 'directory',
-      mode    => '0644',
       recurse => true
     }
 
     if $elasticsearch::params::pid_dir {
       file { $elasticsearch::params::pid_dir:
         ensure  => 'directory',
-        mode    => '0644',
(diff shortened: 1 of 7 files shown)


jordansissel:
lgtm

jordansissel:
(didn't test, but specs look good and puppet manifests look good)

manifests/config.pp
```diff
@@ -48,9 +48,30 @@
       mode   => '0644'
     }
 
+    file { $elasticsearch::params::logdir:
+      ensure  => 'directory',
+      mode    => '0644',
+      recurse => true
+    }
+
+    file { $elasticsearch::params::homedir:
```
jordansissel: We discussed this over video; it'll change bin/elasticsearch's permissions. Richard's working on a fix!

jordansissel:
Last two commits - LGTM

PR #229, opened 2015-01-11
add systemd support for RHEL 7

yanivpaz:
Change code to include systemd option :

[ 'init','systemd'  ]

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -137,7 +137,7 @@ class elasticsearch::params {
       $service_hasrestart = true
       $service_hasstatus  = true
       $service_pattern    = $service_name
-      $service_providers  = [ 'init' ]
+      $service_providers  = [ 'init','systemd'  ]
       $defaults_location  = '/etc/sysconfig'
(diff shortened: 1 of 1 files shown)


electrical:
Hi,

Thanks for the PR.
It seems time to refactor that services part to better select which service provider is required for which linux distro and version.
I don't think people like to set it manually :-)
Think it makes sense to select systemd if its centos/RHEL 7 and sysv init for the older ones?

yanivpaz:
Hi
Yes i think that the default to el7 should be changed to sysyemd.
Anyway , if someone wants to use initv he can set it in hiera file
Thanks
Yaniv

electrical:
Hi,

I've improved the handling of systemd and distro's that use it via PR https://github.com/elasticsearch/puppet-elasticsearch/pull/258
I hope this will work for you?

yanivpaz:
Thanks alot !!we will test it and update .

electrical:
Thanks! i'll close this PR. if you run into any issues can you open a new issue? thanks!

PR #254, opened 2015-01-28
Remove port check as it caused some weird issues at times

electrical:


electrical:
Failure unrelated to change. Single PE run failed. Most likely host is overloaded.

PR #264, opened 2015-02-08
update dependencies

cristi1979:
update dependencies

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -12,6 +12,18 @@
     {
       "name": "puppetlabs/stdlib",
       "version_requirement": ">= 3.2.0 < 5.0.0"
+    },
+    {
+      "name": "puppetlabs/apt",
+      "version_requirement": "1.x"
+    },
+    {
+      "name": "puppetlabs/java",
+      "version_requirement": "1.x"
+    },
+    {
+      "name": "darin/zypprepo",
+      "version_requirement": "1.x"
     }
   ],
   "operatingsystem_support": [
--- a/metadata.json
+++ b/metadata.json
@@ -15,15 +15,15 @@
     },
     {
       "name": "puppetlabs/apt",
-      "version_requirement": "1.x"
+      "version_requirement": ">= 1.0.0 < 2.0.0"
     },
     {
       "name": "puppetlabs/java",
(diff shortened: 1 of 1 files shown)


electrical:
Hi,

Reason i haven't included these dependencies by default is because they are optional and very dependant on the systems you want to run it on.
I wanted to avoid having modules being installed while they may not be used.

Any opinions on that?

cristi1979:
From my point of view, it's not a correct approach. The default installation should work regardless of what parameters are used. 

Especially with hiera usage, the parameters can be changed without touching any puppet code.

If someone wants a cleaner environment, they should be the ones to decide what to remove.

electrical:
I see your point.
I'll give this some more thought and have an internal discussion about it and will get back to you.
Will most likely be over a week since upcoming week i'm off.

electrical:
I've decided to go with specifying the dependencies in the metadata.
Could you change the version specifications to what the stdlib is like?

for example for the java module into `">= 1.3.0 < 2.0.0"`

thanks!

cristi1979:
I've updated the deps versions.

I would like to let you know that I opened similar pull request for other modules and those have been rejected. Those are the links if you want to read the rationals they offered:
    https://github.com/theforeman/puppet-puppet/pull/235#issuecomment-73472941
    https://github.com/jenkinsci/puppet-jenkins/pull/246#issuecomment-73409859

Just for your information. We currently keep those updated manually on our side.

electrical:
@cristi1979 thank you for the references.

Gave it some more thought and looked at the arguments given.
Since the dependencies are classified as soft-dependencies and are only used in certain situations I don't think i would want them added by default.
Also the argument of people possibly using other modules to manage java or apt would mean conflicts indeed if i would force them to install the ones i want.
Also for the past 2 years that the module exists this has never been an issue so i'm a bit reluctant to start forcing it now.

But i do understand the other side of the coin that it could create confusion if certain modules are not installed which might be required for the module to work correctly.
I think as long as all soft dependencies are clearly wrote down it should at least give a starting point for most people.

cristi1979:
Ok then I will close this PR. Thank you.

PR #271, opened 2015-02-18
Set undefined variables to work with strict_variables

halfninja:
Some changes to avoid errors when strict variables are enabled.

Would previously fail when removing any instance, and when creating an instance on CentOS.

Assuming `service_hasstatus` should be false as it would have interpreted `undef` as such, but if the RedHat/CentOS service does support status then that should be true.

### DIFF
--- a/manifests/instance.pp
+++ b/manifests/instance.pp
@@ -278,6 +278,8 @@ define elasticsearch::instance(
 
     $require_service = undef
     $before_service  = File[$instance_configdir]
+
+    $init_defaults_new = {}
   }
 
   elasticsearch::service { $name:
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -147,6 +147,7 @@ class elasticsearch::params {
 
       $service_name       = 'elasticsearch'
       $service_hasrestart = true
+      $service_hasstatus  = false
       $service_pattern    = $service_name
       $defaults_location  = '/etc/sysconfig'
       $pid_dir            = '/var/run/elasticsearch'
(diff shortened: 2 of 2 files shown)


halfninja:
I also tried reproducing it in the specs but wasn't having much luck with it (though it did display some similar bugs with strict variables in the specs themselves)

electrical:
Thank you very much for the PR and testing out the strict variables part.
Regarding the `service_hasstatus` part it seems i messed up at c048c72dae5fa0068ac4df785a0077f99da8e88d and accidentally removed the setting.
Could you modify it to true?

Currently i haven't ran the spec tests with strict variables set but its in the planning to do so.

halfninja:
No problem, have pushed the change to `true`.

electrical:
Thanks! would you be able to sign the [CLA](http://www.elasticsearch.org/contributor-agreement/) so i can merge it?

halfninja:
I have done already, does it take a few minutes to kick in?

electrical:
Ah okay, sorry. can take a bit before the other systems see it indeed. will keep an eye on it.

Again thank you for the patch!

electrical:
Did some spec testing with future parser and the unassigned variables never showed up as a source of failure. ( do have some other failures popping up )
Did you find it out by actually running with the future parser enabled?

halfninja:
No problem, thanks for looking at the PR so promptly.

Yep, am actually running with parser=future and strict_variables=true which is where I saw the errors. I couldn't reproduce it with the specs either but I was putting that down to my inability to drive rspec properly.

electrical:
I haven't been able to reproduce it either with future parser and strict variables enabled with the spec runs so its not just you. haha.

electrical:
thank you very much for the PR. it will be included in the next release ( which i hope to do soon-ish )

Issue #275, opened 2015-02-24
Create a connection validator resource


Spredzy:
Some Puppet modules (mongodb[1], puppetdb[2]) do provide a X_conn_validator resource, where X is the name of the service to check. This make it possible to "suspend" (while validating) the puppet run, until the service is actually up and running.

Elasticsearch bootstrap is not instantaneous, hence currently, it is possible to apply resources that depends on the Elasticsearch instance to be up and running, but that is not (even with ordering).
## Proposed solution
- First patchset : create an `es_instance_conn_validator` resource that would check that port P is listening on host H, applying the same pattern as mongodb and puppetdb.

``` puppet
es_instance_conn_validator { 'myinstance' :
  server  => 'foo.bar',
  port    => '9200',
  timeout => 60,
}
```
- Second patchset : create a `$validate_instance` parameter for `elasticsearch::instance`, that if enabled will integrate the `es_instance_conn_validator_resource` in the catalog for that instance.

[1] https://github.com/puppetlabs/puppetlabs-mongodb/blob/master/lib/puppet/type/mongodb_conn_validator.rb
[2] https://github.com/puppetlabs/puppetlabs-puppetdb/blob/master/lib/puppet/type/puppetdb_conn_validator.rb

electrical:
The first patchset hasb een merged.
Feel free to send a PR for the second one.
Cheers.

electrical:
Hiya, will you be having time to work on this further? Worst case i'll leave this open for some one else to implement or when i have time.

Spredzy:
Heya @electrical, sorry for the lack of activity on this one. Unfortunately I won't have time in the near future to work on that issue, I agree on leaving it open if anyone wants to take it.

PR #276, opened 2015-02-24
es_instance_conn_validator: Resource to verify instance availability

Spredzy:
This commit brings a new resource `es_instance_conn_validator` that
aims to validate that an instance is up & running.

Fix: #275

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,6 +202,25 @@ elasticsearch::python { 'rawes': }
 elasticsearch::ruby { 'elasticsearch': }
 ```
 
+###Connection Validator
+
+This module offers a way to make sure an instance has been started and is up and running before
+doing a next action. This is done via the use of the `es_instance_conn_validator` resource.
+```puppet
+es_instance_conn_validator { 'myinstance' :
+  server => 'es.example.com',
+  port   => '9200',
+}
+```
+
+A common use would be for example :
+
+```puppet
+class { 'kibana4' :
+  require => Es_Instance_Conn_Validator['myinstance'],
+}
+```
+
 ###Package installation
 
 There are 2 different ways of installing the software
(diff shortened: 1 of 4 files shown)


electrical:
@Spredzy looks great!
Is it possible to have the host default to 'localhost' ? That's assuming the resource runs from the node it self of course.

Could you also add an example to the readme for how to use it in most cases?
I would assume a require to it to ensure its up?

lib/puppet/type/es_instance_conn_validator.rb
```diff
@@ -0,0 +1,38 @@
...
+    defaultto 'localhost'
+  end
+
+  newparam(:port) do
+    desc 'The port that the Elasticsearch instance should be listening on.'
+    defaultto 9200
+  end
+
+  newparam(:timeout) do
+    desc 'The max number of seconds that the validator should wait before giving up and deciding that puppetdb is not running; defaults to 60 seconds.'
```
electrical: s/puppetdb/elasticsearch/

README.md
```diff
@@ -202,6 +202,25 @@ elasticsearch::python { 'rawes': }
 elasticsearch::ruby { 'elasticsearch': }
 ```
 
+###Connection Validator
+
+This module offers a way to make sure an instance has been started and is up and running before
+doing a next action. This is done via the use of the `es_instance_conn_validator` resource.
+```puppet
+es_conn_validator { 'myinstance' :
```
electrical: Should be `es_instance_conn_validator` i think?

README.md
```diff
@@ -202,6 +202,25 @@ elasticsearch::python { 'rawes': }
...
+  server => 'es.example.com',
+  port   => '9200',
+}
+```
+
+A common use would be for example :
+
+```puppet
+class { 'kibana4' :
+  require => Es_Conn_Validator['myinstance'],
```
electrical: same issue as above.

PR #288, opened 2015-03-06
change log's dir

xiaochou:


electrical:
Hi,

changing the log directory takes a few more things then just a variable.
Main reason is that all the init script define the log directory as well so it means we need to change it in the services handling to modify it.

Could you also write a spec test for this? With a default value and custom value to ensure we change all the parts correctly?

Thanks!

electrical:
Closing due to no reply

PR #294, opened 2015-03-20
Update apt key to full 40characters

exptom:
Latest version of puppetlabs/apt module shows warning on every puppet run if using a short key so this update includes the full 40 character key.

### DIFF
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -40,7 +40,7 @@ class elasticsearch::repo {
         location    => "http://packages.elasticsearch.org/elasticsearch/${elasticsearch::repo_version}/debian",
         release     => 'stable',
         repos       => 'main',
-        key         => 'D88E42B4',
+        key         => '46095ACC8548582C1A2699A9D27D666CD88E42B4',
         key_source  => 'http://packages.elasticsearch.org/GPG-KEY-elasticsearch',
         include_src => false,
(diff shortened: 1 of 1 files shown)


electrical:
Strange that i haven't seen this yet with any of my tests.
Will do some checking on it.

electrical:
@exptom would you be able to sign the [CLA](http://www.elastic.co/contributor-agreement/) so i can merge it? Thanks!

exptom:
@electrical Done! :)

electrical:
jenkins, test it

electrical:
@exptom thank you for the PR. Tests are green. Merging.

electrical:
Hi,

I'm very sorry but i had to revert this PR.
It causes issues with some older versions of the apt module that are shipped with some older Puppet Enterprise versions which i still support.
At this moment i prefer to have them still working while newer apt module versions  just to alert about it.
When i drop support for those older versions i can revisit this.

onibox:
@electrical What about using $::is_pe from puppetlabs stdlib to distinguish between Puppet Enterprise and Puppet Open Source? See https://github.com/elastic/puppet-elasticsearch/pull/607

cdenneen:
Newer PE would break as well as OS. So this would need to be PE and puppetversion <= X.Y

PR #303, opened 2015-03-29
Set PATH from facter to resolve #286

### LINKED ISSUE #286: Specifying JAVA_HOME in init defaults is not exported in init.d script
Specifying JAVA_HOME in init defaults is not exported in RedHat init.d script. 
Add a line to check for definition of JAVA_HOME and export it after sourcing the defaults file.

flopex:
This will set the ```PATH``` in ```/etc/sysconfig/elasticsearch-instancename``` used for the System V init script ```elasticsearch-instancename``` to be what ```facter``` has. This is to mitigate the issue where running ```service elasticsearch-instancename``` does not recognize the system's PATH. As a more permanent fix you could source ```/etc/profile.d``` to better emulate the the user's environment.

Tested on: CentOS 6.6 x64

### DIFF
--- a/manifests/instance.pp
+++ b/manifests/instance.pp
@@ -256,7 +256,7 @@ define elasticsearch::instance(
       $global_init_defaults = { }
     }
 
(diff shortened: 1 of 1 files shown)


electrical:

electrical:
Hi,

Which specific issue does this solve?
None of the acceptance tests have exposed any issues nor any other reported issues so far.
Thanks!

PR #309, opened 2015-04-01
Add the ability to create and populate scripts

electrical:
- Adding general scripts directory where they will be placed
- Adding scripts directory per instance which symlinks to the general directory
- Adding spec tests and acceptance tests

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -98,6 +98,11 @@ class elasticsearch::config {
       require => [ Exec['mkdir_templates_elasticsearch'] ]
     }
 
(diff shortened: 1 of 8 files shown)


drewr:
Seems straightforward enough. The user is expected to provide the linked-to dir? Or will it look like something else?

afx114:
I think this looks good, the one thing I would change is to ensure that the scripts dir is also owned by the elasticsearch user.

electrical:
@drewr the link to dir is automatically done by the main class.

@afx114 good point. Will take care of that.

PR #313, opened 2015-04-09
LimitNOFILE and LimitMEMLOCK in elasticsearch.systemd.erb

Filirom1:
Hi,

With systemd, nofile and memlock limits are not configurable.
This Pull Request add support for custom LimitMEMLOCK and LimitNOFILE in elasticsearch.systemd.erb

Cheers
Romain

### DIFF
--- a/manifests/service/systemd.pp
+++ b/manifests/service/systemd.pp
@@ -151,6 +151,18 @@ define elasticsearch::service::systemd(
       $pid_dir           = $elasticsearch::pid_dir
       $defaults_location = $elasticsearch::defaults_location
 
+      if ($new_init_defaults != undef and is_hash($new_init_defaults) and has_key($new_init_defaults, 'MAX_OPEN_FILES')) {
+        $nofile = $new_init_defaults['MAX_OPEN_FILES']
+      }else{
+        $nofile = '65535'
+      }
+
+      if ($new_init_defaults != undef and is_hash($new_init_defaults) and has_key($new_init_defaults, 'MAX_LOCKED_MEMORY')) {
+        $memlock = $new_init_defaults['MAX_LOCKED_MEMORY']
+      }else{
+        $memlock = undef
+      }
+
       file { "/lib/systemd/system/elasticsearch-${name}.service":
         ensure  => $ensure,
         content => template($init_template),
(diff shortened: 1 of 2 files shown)


manifests/service/systemd.pp
```diff
@@ -149,6 +149,14 @@
       $user  = $elasticsearch::elasticsearch_user
       $group = $elasticsearch::elasticsearch_group
 
+      if ($new_init_defaults != undef) {
```
electrical: this variable will never be undef.
Its best to check if the keys actually exist before using them.

templates/etc/init.d/elasticsearch.systemd.erb
```diff
@@ -10,9 +10,11 @@ Group=<%= @group %>
 PIDFile=/var/run/elasticsearch/elasticsearch-<%= @name %>.pid
 ExecStart=/usr/share/elasticsearch/bin/elasticsearch -d -p /var/run/elasticsearch/elasticsearch-<%= @name %>.pid -Des.default.config=$CONF_FILE -Des.default.path.home=$ES_HOME -Des.default.path.logs=$LOG_DIR -Des.default.path.data=$DATA_DIR -Des.default.path.work=$WORK_DIR -Des.default.path.conf=$CONF_DIR
 # See MAX_OPEN_FILES in sysconfig
-LimitNOFILE=65535
+LimitNOFILE=<%= @nofile %>
 # See MAX_LOCKED_MEMORY in sysconfig, use "infinity" when MAX_LOCKED_MEMORY=unlimited and using bootstrap.mlockall: true
-#LimitMEMLOCK=infinity
+<% if @memlock == 'unlimited' %>
+LimitMEMLOCK=infinity
+<% end %>
```
electrical: In case someone doesn't want to set it to infinity perhaps we should add an else statement to use the actual value?
If that would never happen we would need to add some kind of validation of MAX_LOCKED_MEMORY value.

electrical:
Sorry for not re-reviewing this sooner. Left 1 comment. thank you so much for the work.

electrical:
@Filirom1 Are you able to address the comment and do a rebase?
Otherwise i can take it over because i would like to have this feature in.

Filirom1:
Sorry for the delay, I added the else statement, and did the rebase

electrical:
thanks! for some reason i still can't merge it.
Are you sure you rebased it?

Filirom1:
Haha, I rebased my 3 commits into one, but I forget to rebase my commits with the master branch.

Filirom1:
It should be better now

electrical:
Much better now. thank you very much!

PR #314, opened 2015-04-10
Provide a means to override apt-key key.

brentclark:
Provide a means to override apt-key key.

Else you get:
Warning: /Apt_key[Add key: D88E42B4 from Apt::Source elasticsearch]: The id should be a full fingerprint (40 characters), see README.

Puppetlabs-apt ( >= v1.8.0)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -159,6 +159,9 @@
 #   Define plugins via a hash. This is mainly used with Hiera's auto binding
 #   Defaults to: undef
 #
+# [*apt_key*]
+#   Override GPG key for Apt, provided by this module.
+#
 # [*plugins_hiera_merge*]
 #   Enable Hiera's merging function for the plugins
 #   Defaults to: false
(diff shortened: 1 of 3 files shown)


electrical:
Hi,

This is something i rather not do and just fix the actual issue by setting the long key instead of the short one via #294

brentclark:
No problem. Thanks for reviewing my contribution.

Maybe add to your dependancy list,  Puppetlabs-apt ( >= v1.8.0).
If I understand correctly, it is something recently the puppetlabs-apt module is "better handling".

electrical:
I wish it would be that easy but with some Puppet Enterprise versions older versions of the apt module are delivered and may cause some issues if they don't upgrade ( or are not able to )
That's why the version range is fairly wide.

electrical:
Also since its an optional dependency ( not enforced via the module ) I leave it up to the user.
Someone who doesn't use any ubuntu/debian machines shouldn't need it :-)

electrical:
@brentclark that said i do want to thank you for your contribution :-)

PR #323, opened 2015-04-28
Use systemd to start service when on debian 8

mfournier:
I tested this on EL7, debian 7 and debian 8. No changes are made on the 2 former platforms.

A couple of things to keep in mind:
 * as discussed on IRC, adding the support for debian8 to the test suite is out of scope, so currently there is no test coverage for this.
 * if a previous version of the module is used to setup elasticsearch on a debian 8 machine, systemd's LSB wrapper will take care of the service *once systemctl daemon-reload* has been run (which this module doesn't do before this patch, so someting else has to trigger this command)
 * if upgrading from a previous version of this module, `/etc/init.d/elasticsearch-xyz` will be left behind (unmanaged & undeleted)

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -159,17 +159,15 @@ class elasticsearch::params {
       $service_pattern    = $service_name
       $defaults_location  = '/etc/default'
 
-      case $::operatingsystemmajrelease {
-        '8': {
-          $init_template     = 'elasticsearch.systemd.erb'
-          $service_providers = 'systemd'
-          $pid_dir           = '/var/run/elasticsearch'
-        }
-        default: {
-          $init_template     = 'elasticsearch.Debian.erb'
-          $service_providers = [ 'init' ]
-          $pid_dir           = false
-        }
+      if (($::operatingsystem == 'Debian' and $::operatingsystemmajrelease >= 8) or
+          ($::operatingsystem == 'Ubuntu' and $::operatingsystemmajrelease >= 15)) {
+        $init_template     = 'elasticsearch.systemd.erb'
+        $service_providers = 'systemd'
+        $pid_dir           = '/var/run/elasticsearch'
+      } else {
+        $init_template     = 'elasticsearch.Debian.erb'
+        $service_providers = [ 'init' ]
+        $pid_dir           = false
       }
     }
     'Darwin': {
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -133,6 +133,12 @@ class elasticsearch::params {
   # service parameters
   case $::operatingsystem {
     'RedHat', 'CentOS', 'Fedora', 'Scientific', 'Amazon', 'OracleLinux', 'SLC': {
+      $service_name       = 'elasticsearch'
+      $service_hasrestart = true
+      $service_hasstatus  = true
+      $service_pattern    = $service_name
+      $defaults_location  = '/etc/sysconfig'
+      $pid_dir            = '/var/run/elasticsearch'
 
       case $::operatingsystemmajrelease {
         '7': {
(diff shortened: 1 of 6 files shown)


electrical:
As discussed on IRC, i'll enable the tests as soon as i have a working Debian 8 docker image so i can actually test the changes :-)

electrical:
Been successful in creating a debian 8 docker image.
Running the PR tests manually now.

manifests/params.pp
```diff
@@ -145,22 +151,26 @@
...
       $service_hasrestart = true
       $service_hasstatus  = true
       $service_pattern    = $service_name
-      $service_providers  = 'init'
       $defaults_location  = '/etc/default'
-      $init_template      = 'elasticsearch.Debian.erb'
-      $pid_dir            = false
+
+      case $::operatingsystemmajrelease {
+        '8': {
```
sathieu: I would go for:
```
if ($::operatingsystem == 'Debian' and versioncmp($::operatingsystemrelease, '8.0') >= 0) or ($::operatingsystem == 'Ubuntu' and versioncmp($::operatingsystemrelease, '15.04') >= 0) {
```
instead. it matches Ubuntu and future releases.

electrical: I think Ubuntu will have to get a separate block so we can make the differences more clearly.
I hate to have those long if statements, especially if they have an 'or' case.
it invites to more horrible stuff :-)

electrical: @mfournier could we change this to everything lower then 8 use init and 8 or above use systemd?
Makes it easier to maintain for the future.

Thanks!

mfournier: Sure! Currently AFK. I'll take care of this tomorrow.

electrical: Great thanks!

mfournier:
Thanks for the review @sathieu & @electrical !

I just updated the patchset to properly support ubuntu.

manifests/params.pp
```diff
@@ -145,22 +151,24 @@
...
       $service_name       = 'elasticsearch'
       $service_hasrestart = true
       $service_hasstatus  = true
       $service_pattern    = $service_name
-      $service_providers  = 'init'
       $defaults_location  = '/etc/default'
-      $init_template      = 'elasticsearch.Debian.erb'
-      $pid_dir            = false
+
+      if (($::operatingsystem == 'Debian' and $::operatingsystemmajrelease >= 8) or
```
electrical: Like i mentioned i would love to see this more verbose instead of a one liner.
My best idea would be to have Debian and Ubuntu have their own spot in the case `$::operatingsystem` case part and then do a check for the `$::operatingsystemmajrelease` value.

thanks!

mfournier:
@electrical here you go... :smile:
Sorry for the misunderstanding ! I aimed for avoiding duplicate statements over code clarity.

electrical:
@mfournier thanks! i do agree that less code duplication would be nice, but sometimes having things explicit makes it easier to manage in the future.

electrical:
jenkins, test it

electrical:
@mfournier since you changed the path of where the service files are placed the rspec tests fail, could you fix that ? ;-)

mfournier:
Here you are @electrical. I just grepped through the spec directory but didn't actually run the tests. Hope I didn't miss any !

jenkins, test it (although I probably don't have the permission to invoke the bot...)

electrical:
Jenkins, test it

electrical:
jenkins, test it

electrical:
@mfournier there seems to be still 1 error. Looks like you missed https://github.com/elastic/puppet-elasticsearch/blob/master/manifests/config.pp#L110 to change the path as well.

Cheers.

mfournier:
These tests appear to be worth the effort, great :-)

Fixed by d1c07e856. I hope we're OK now, @electrical.

electrical:
jenkins, test it

manifests/params.pp
```diff
@@ -145,22 +151,40 @@
...
+    'Debian': {
       $service_name       = 'elasticsearch'
       $service_hasrestart = true
       $service_hasstatus  = true
       $service_pattern    = $service_name
-      $defaults_location  = '/etc/sysconfig'
-      $pid_dir            = '/var/run/elasticsearch'
+      $defaults_location  = '/etc/default'
+
+      if ($::operatingsystemmajrelease >= 8) {
```
electrical: Forgot i enabled [this](https://github.com/puppet-community/puppet-lint-version_comparison-check) check which causes puppet-lint now to fail.
See https://github.com/puppet-community/puppet-lint-version_comparison-check#version-compared-as-number for details.

electrical:
@mfournier i'm merging it as is and will fix any other issues. Thank you so much for the contribution!

mfournier:
Wonderful, thanks !

PR #324, opened 2015-04-29
puppetlabs/apt 2.0.x compatibility

zsim0n:
Hello,
Here is my version for puppetlabs/apt 2.0.x compatibility. I have tested on : 
Mac OS X 10.10.3 , vagrant 1.7.2, virtualbox 4.3.26, ubuntu 14.04, puppet 3.7.5 and librarian-puppet 2.1.0

Best
Zoltan


electrical:
Hi,

I'll be pinning the module to use apt < 2.0

PR #326, opened 2015-05-01
Updated apt::source. Now should work with puppetlabs-apt versions 2.x

alejandrobednarik-olx:


electrical:
Hi,

I'll be pinning the module to use apt < 2.0

PR #337, opened 2015-05-21
fix for puppetlabs-apt 2.0 

Tombar:
Hello,

Just a small PR to support puppetlabs apt 2.0 changes + a exec apt-get update before package install.

This PR would break backward compatibility with older versions of puppetlabs-apt

Regards

M


electrical:
Hi,

In the readme i specified for now that I'm only supporting pl-apt 1.8 and lower.
When 1.0 is out i will revise it.

PR #341, opened 2015-05-24
Added support for puppetlabs-apt >= 2.0

zdenekjanda:
This is needed for puppetlabs-apt >= 2.0 support.


electrical:
Hi,

In the readme i specified for now that I'm only supporting pl-apt 1.8 and lower.
When 1.0 is out i will revise it.

PR #346, opened 2015-05-28
Add automatic plugin directory extraction

electrical:
The function plugin_dir extracts the resulting end directory name that
will be used by Elasticsearch where the plugin will be installed in.
Added some tests as well and moved the facts spec to the same location


electrical:
jenkins, test it

electrical:
At the moment the pluginmanager code handles elasticsearch and elastic as first name so i kept to the same part.
haven't seen anything else named with a different prefix name so far.

philipwigg:
Do you mean inside the parse function we were looking at before? I see elasticsearch- and es- (but not elastic-) ?

electrical:
I must be going nuts :-)  lol
Will change it.

electrical:
Still need to refactor the plugin define to make use of it to make sure it works like intended.
Also need to deprecate the current `module_dir` option to make it do nothing. Maybe just log a warning.

electrical:
Deprecation notice would need to be logged with a [warning](https://docs.puppetlabs.com/references/latest/function.html#warning)

electrical:
@philipwigg could you do a quick check if its correct? spec and unit testing passes so i assume its okay :-)

philipwigg:
Looks great! This will make the plugin feature much easier to use I think.

The only suggestion I would make is that having the default remove setting for the plugin define could be a bit dangerous. Maybe make it so that you have to put 'absent' or whatever and it just fails with if there's an unrecognised value.

manifests/plugin.pp
```diff
@@ -129,8 +133,8 @@
     }
     default: {
       exec {"remove_plugin_${name}":
-        command => "${elasticsearch::plugintool} --remove ${module_dir}",
-        onlyif  => "test -d ${elasticsearch::plugindir}/${module_dir}",
+        command => "${elasticsearch::plugintool} --remove ${plugin_dir}",
+        onlyif  => "test -d ${elasticsearch::plugindir}/${plugin_dir}",
         notify  => $notify_service,
       }
     }
```
philipwigg: This could be a bit dangerous having this as the default. I realise it's not changed in this commit but thought I'd mention it.

electrical:
Good catch! should have changed that ages ago. will add that change and push it then.
thanks!

electrical:
@philipwigg that issue should be solved now as well :-)

electrical:
jenkins, test it

electrical:
jenkins, test it

Issue #333, opened 2015-05-12
elasticsearch module not compatible with new puppetlabs/apt module


TravellingGuy:
The latest puppetlabs/apt module (v. 2.0.0) has had a major refactor.

When the elasticsearch module is configured to manage the repo (manage_repo => true), puppet agent fails with an error:
> Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Invalid parameter key_source

philipwigg:
Hi, this is a duplicate of #322.

Basically you need to use the apt module < 2.0 and the next release will enforce that.

The module can't support apt 2.0 yet because it breaks compatibility with older PE versions.

TravellingGuy:
Thanks. I can live with the incompatibility for the time being.

Incidentally, I did do a search for the problem before posting my new issue, but I neglected to search Closed issues. Doh.

Issue #322, opened 2015-04-24
Installation broken for puppetlabs-apt 2.0.0


claussni:
The repo installation doesn't work with the recent `puppetlabs-apt 2.0.0`. The error is:
```
Error: Invalid parameter key_source on Apt::Source[elasticsearch] at /etc/puppet/modules/elasticsearch/manifests/repo.pp:46
```
Here is the module list:
```
/etc/puppet/modules
├── elasticsearch-elasticsearch (v0.9.5)
├── puppetlabs-apt (v2.0.0)
├── puppetlabs-java (v1.3.0)
└── puppetlabs-stdlib (v4.6.0)
```
It works again by forcing version 1.8.0 of `puppetlabs-apt`.

[This commit](https://github.com/puppetlabs/puppetlabs-apt/commit/0f3bdcdf5a44315197ded803401b52dfcce38ebe) in puppetlabs-apt seems to break it.

electrical:
I noticed the same issue today with the spec tests indeed.
Next release will force to be below 2.0 so it won't break it for the time being.

alejandrobednarik-olx:
I already made a pull request to fix this issue.

claussni:
For the sake of documentation: https://github.com/elastic/puppet-elasticsearch/pull/326

electrical:
I've done a documentation update stating apt < 2.0 is required.
Closing this issue.

madAndroid:
This should be resolved with the latest release of the `puppetlabs-apt` module - see https://github.com/puppetlabs/puppetlabs-apt/blob/master/CHANGELOG.md#summary

PR #347, opened 2015-05-29
APT 2.0 support

s12v:
Adds support for puppetlabs-apt 2.0

`bundle exec rake spec` and `bundle exec rake lint` finished without errors.

I couldn't run beaker tests:
```
Beaker::Hypervisor, found some docker boxes to create
Provisioning docker
provisioning ubuntu-14-04
Using docker server at 192.168.59.103
Warning: Try 1 -- Host 192.168.59.103 unreachable: Errno::ECONNREFUSED - Connection refused - connect(2) for "192.168.59.103" port 32774
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ This module has been tested against ES 1.0 and up.
 #### Repository management
 When using the repository management you will need the following dependency modules:
 
-* Debian/Ubuntu: [Puppetlabs/apt](http://forge.puppetlabs.com/puppetlabs/apt) Version 1.8.x or lower.
+* Debian/Ubuntu: [Puppetlabs/apt](http://forge.puppetlabs.com/puppetlabs/apt)
 * OpenSuSE: [Darin/zypprepo](https://forge.puppetlabs.com/darin/zypprepo)
 
 ##Usage
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -36,13 +36,27 @@ class elasticsearch::repo {
         class { 'apt': }
       }
 
-      apt::source { 'elasticsearch':
-        location    => "http://packages.elasticsearch.org/elasticsearch/${elasticsearch::repo_version}/debian",
-        release     => 'stable',
-        repos       => 'main',
-        key         => 'D88E42B4',
-        key_source  => 'http://packages.elasticsearch.org/GPG-KEY-elasticsearch',
-        include_src => false,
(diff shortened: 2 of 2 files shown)


philipwigg:
I don't think there are any plans to fix this at the moment because older PE versions that the module needs to support are pinned to older puppetlabs-apt versions.

See https://github.com/elastic/puppet-elasticsearch/issues/322 and https://github.com/elastic/puppet-elasticsearch/issues/333.

@electrical, do you think we can re-open one of these issues so that it's visible by default in the issues lilst? Will hopefully prevent any more people submitting PRs.

s12v:
I see. But with this change it will support both 1.x and 2.x

electrical:
jenkins, test it

s12v:
@electrical, looks like puppet 3.2.0 doesn't work with apt 2.0, I've reverted changes in `.fixtures.yml` and `spec_helper_acceptance.rb` (so tests will run with apt 1.8)

electrical:
@s12v Kinda defeats the purpose right? Also acceptance testing is not corrected to make sure things keep on working with this change.
Because testing wise we are not ready Im not confident in merging this.
Will take an other look at it Monday and see what direction we can go for this.

s12v:
@electrical, well, usually tests don't care about module versions, so anyway all possible cases will not be covered. Not sure if it's possible to have different module versions in `.fixtures.yml` based on `PUPPET_VERSION`?

But I'd really like to try the acceptance tests (because for another puppet module apt 2.0 didn't do `Exec[apt_get]` after adding a key (unlike 1.8), so it couldn't install; not sure how it would go with your module). Do you have any idea what the error from my first comment means?

s12v:
@electrical, tested it in vagrant and both 1.8 and 2.0 seem to work: https://gist.github.com/s12v/946538bc45a01b1bb962

electrical:

madAndroid:
Regarding this change: a newer version of `puppetlabs-apt` (2.1.0) is backwardly compatible with 1.8.x - see https://github.com/puppetlabs/puppetlabs-apt/blob/master/CHANGELOG.md#summary

liamjbennett:
Any further update on this? We should be able to update the metadata.json dependency now to support the 2.1.0 release for those people using it.

electrical:
Hi,

#452 was merged to solve this issue and should work.
I'll be bumping the version in the metadata file as well.

PR #356, opened 2015-06-02
use init service provider for Amazon Linux

alekseymykhailov:
release 0.9.6 breaks (https://github.com/elastic/puppet-elasticsearch/commit/066f70c99d7010ec9ddfe07ec87f4671765c68c4) service management on Amazon Linux since it does not have systemd installed
this change creates separate case for Amazon Linux to allow to define "init" as service provider

facter output from AWS linux

```
# facter  operatingsystem
Amazon
# facter  operatingsystemmajrelease
2015
```

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -132,7 +132,7 @@ class elasticsearch::params {
 
   # service parameters
   case $::operatingsystem {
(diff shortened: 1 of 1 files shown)


electrical:
Thank you for re-doing the commit.
I'll let the test system do its thing and if everything is green i'll merge it.
I'm planning to do a release next week.
Thanks!

electrical:
jenkins, test it

electrical:
Thank you again for the patch.

PR #366, opened 2015-06-10
Fix systemd template

### LINKED ISSUE #362: default behavior on centos7 breaks unit file
if the init_defaults param for an instance resource doesn't  have a MAX_LOCKED_MEMORY key, the LimitMEMLOCK= varaible in the unit file will be empty, resulting in these error messages:
```
Jun 10 10:11:02 clusternode1..de systemd[1]: [/usr/lib/systemd/system/elasticsearch-es-01.service:16] Failed to parse resource value, ignoring:
```

the unit file:
```
[Unit]
Description=Starts and stops a single elasticsearch instance on this system
Documentation=http://www.elasticsearch.org

[Service]
Type=forking
EnvironmentFile=/etc/sysconfig/elasticsearch-es-01
User=elasticsearch
Group=elasticsearch
PIDFile=/var/run/elasticsearch/elasticsearch-es-01.pid
ExecStart=/usr/share/elasticsearch/bin/elasticsearch -d -p /var/run/elasticsearch/elasticsearch-es-01.pid -Des.default.config=$CONF_FILE -Des.default.path.home=$ES_HOME -Des.default.path.logs=$LOG_DIR -Des.default.path.data=$DATA_DIR -Des.default.path.work=$WORK_DIR -Des.default.path.conf=$CONF_DIR
# See MAX_OPEN_FILES in sysconfig
LimitNOFILE=65535
# See MAX_LOCKED_MEMORY in sysconfig, use "infinity" when MAX_LOCKED_MEMORY=unlimited and using bootstrap.mlockall: true

LimitMEMLOCK=

# Shutdown delay in seconds, before process is tried to be killed with KILL (if configured)
TimeoutStopSec=20

[Install]
WantedBy=multi-user.target
```

my profile:
```
class profiles::elasticsearch {
  class { '::elasticsearch':
    repo_version  => '1.6',
    java_install  => true,
    manage_repo   => true,
    config        => { 'cluster.name' => 'elasticsearch_cluster_1' },
    elasticsearch_user  => 'elasticsearch',
    elasticsearch_group => 'elasticsearch',
    autoupgrade        => true,
  } ~>
  elasticsearch::instance { 'es-01': 
    config  => { 'node.name' => $::hostname },
    init_defaults => {
                        'ES_USER'       => 'elasticsearch', 
                        'ES_GROUP'      => 'elasticsearch', 
                        'user'          => 'elasticsearch', 
                        'group'         => 'elasticsearch',
                        'ES_HEAP_SIZE'  => '16g'},
  }
  elasticsearch::plugin { 'lmenezes/elasticsearch-kopf':
#    ensure      => absent, # set to absent to remove it, than reenable it to get the latest version. currently the only supported update mechanism https://github.com/elastic/puppet-elasticsearch/pull/213
    module_dir  => 'kopf',
    instances   => 'es-01',
    proxy_host  => 'proxy..de',
    proxy_port  => 3128,
  }

```

the variable $memlock defaults to undef in the systemd.pp at line 163, this should be changed to infinity or any suitable int?

electrical:
When MAX_LOCKED_MEMORY was not set it would leave an empty setting in
the systemd file which caused some errors.
Now we check if its actually set ( not undef ) before printing its
value.

Closes #362

### DIFF
--- a/templates/etc/init.d/elasticsearch.systemd.erb
+++ b/templates/etc/init.d/elasticsearch.systemd.erb
@@ -14,7 +14,7 @@ LimitNOFILE=<%= @nofile %>
 # See MAX_LOCKED_MEMORY in sysconfig, use "infinity" when MAX_LOCKED_MEMORY=unlimited and using bootstrap.mlockall: true
 <% if @memlock == 'unlimited' %>
 LimitMEMLOCK=infinity
-<% else %>
+<% elsif @memlock %>
 LimitMEMLOCK=<%= @memlock %>
 <% end %>
 # Shutdown delay in seconds, before process is tried to be killed with KILL (if configured)


bastelfreak:
just tested it on centos7, works fine. +1 for merge

electrical:
Failure due to disk space issues on docker servers :-( Merging anyway since most did pass :-)

PR #375, opened 2015-06-12
Add Support for SLES.

buzzdeee:
This adds support for Novell SLES. Developed and tested on SLES 11 SP3.
Updated README.md and metadata.json to reflect added support.
Template for the init script is very similar to the RedHat one, adapted for SLES.

Please consider for merge. Let me know if there is anything to enhance/change, etc, and I'm happy to do so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ This module has been tested against all versions of ES 1.x and 2.x
 When using the repository management you will need the following dependency modules:
 
 * Debian/Ubuntu: [Puppetlabs/apt](http://forge.puppetlabs.com/puppetlabs/apt)
-* OpenSuSE: [Darin/zypprepo](https://forge.puppetlabs.com/darin/zypprepo)
+* OpenSuSE/SLES: [Darin/zypprepo](https://forge.puppetlabs.com/darin/zypprepo)
 
 ##Usage
 
@@ -517,6 +517,7 @@ The module has been tested on:
 * CentOS 6/7
 * Ubuntu 12.04, 14.04
 * OpenSuSE 13.x
+* SLES 11 SP3
 
 Other distro's that have been reported to work:
 
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -121,7 +121,7 @@ class elasticsearch::params {
       # main application
       $package = [ 'elasticsearch' ]
     }
-    'OpenSuSE': {
+    'OpenSuSE', 'SLES': {
       $package = [ 'elasticsearch' ]
     }
     default: {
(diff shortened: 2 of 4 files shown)


electrical:
Thank you very much for this PR.
I know that on the Elasticsearch side work has been done for making the RedHat init script work with SLES, See https://github.com/elastic/elasticsearch/pull/9366/files
Is it an idea to use that instead of an separate init script?

Also i see the way you handle the service part with `startproc`, i can't think if there was an issue with it or not.
The main reason for merging it into a single init script is that at some point the RPM should work on all RPM based distro's including OpenSuse / SLES.

What are your thoughts about it?

buzzdeee:
I had that idea first too, just using the RedHat script, but I ran into trouble, I just verified and used the RedHat script:

 # /etc/rc.d/elasticsearch-es-01 start
Starting elasticsearch-es-01: /etc/rc.d/elasticsearch-es-01: line 89: daemon: command not found

# which daemon
which: no daemon in (/sbin:/usr/sbin:/usr/local/sbin:/root/bin:/usr/local/bin:/usr/bin:/bin:/usr/X11R6/bin:/usr/games:/usr/lib64/jvm/jre/bin:/usr/lib/mit/bin:/usr/lib/mit/sbin)

therefore I used that startproc approach I know is working.
I guess the other approach used in the script in the PR, is tested on SLES and supposed to work too, so using that instead, would work for me too.
Question is, who is going to do that, me, or you, and then either prior to merging this or afterwards ;)
Let me know.

btw: I currently use elasticsearch-1.5.1-1 rpm, which only brings the RedHat init scrip with the  under /etc/rc.d/init.d/elasticsearch, therefore a working init script is kind of required for the module for me to work with it ;)

cheers,


btw: that CLA I just signed.

electrical:
It would be nicer to have `startproc` then `su -s` i think assuming it works like expected and be able to run in different users.

buzzdeee:
rebased to current master

buzzdeee:
again rebased to current master

electrical:
Hi, sorry for the delay in merging. I want to make sure I'm able to test it with the acceptance test suite so i don't break anything at some point :-)

buzzdeee:
sure, I didn't wanted to rush, just trying to keep up ;)

electrical:

electrical:
Hi,

Been doing some testing and running into issues with `rpmkeys` not existing on SLES 11 SP3. Can you check if that is available via some other package i may not have installed?
Worst case the rpm key import will have to be re-factored.

electrical:
By the looks of it i could replace it to use `rpm --import`  and then give it the url.
Will have to check if this works for OpenSuse as well.

buzzdeee:
for your convenience, rebased, there was a little conflict in README.md

tylerjl:
Changes have been merged into the combined PR in #609, closing to track in that PR.

Issue #363, opened 2015-06-10
split out restart_on_change into more granular options


electrical:
restart_on_change is to generic and should be expanded.
The basis of these options will also be that we choose the most safest possible defaults since ES is very sensitive to random restarts.

Currently thinking of the following options:

## Module generic options

- restart_on_change ( Default: false )
Master control for all other options.

- restart_package_change ( Default: false )
This set to true would restart the service if we upgrade the package.

- restart_config_change ( Default: false )
This set to true would restart the service if we change the configuration.

- restart_service_change ( Default: false )
Controls if we restart or not on any changes related to the services ( Defaults file / init / systemd )

## Es module specifics

- restart_plugin_change ( Default: false )
Set to true would restart the service when a plugin is installed. this is required for any java based plugins.

## Related to

#336

bastelfreak:
Sounds good so far, restart_service_change will only restart the service if something in the servicefile itself changes and restart_config_change for every other configuration file? I'm not sure if it is needed to have two seperate params for these. 

I can't imagine a scenario where I change something in the config file and want to restart ES, but after that change something in the service file and don't want to restart.

electrical:
Very good point.
restart_config and restart_package only then and we should be good?

bastelfreak:
Yes, this should be sufficient.

tylerjl:
Completed as of #655, will be part of the 1.0.0 release.

PR #384, opened 2015-06-20
Restart Elasticsearch only when cluster status is green, 

parabolic:
we have this working in production because we are using puppet for managing
elasticsearch, and on every plugin install we've had to do it
manually(which beats the purpose of puppet)in places where the cluster
has 2 or 3 nodes because  it can happen in some cases that puppet runs at the same
time and the whole cluster goes down since it takes time for one of them
to resync the data and while that is happening the other one restarts.
With this check puppet will wait until all data is synced and allocated properly ( cluster state green)
and then will apply the plugin, restarting the node.

### DIFF
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -146,7 +146,7 @@ define elasticsearch::plugin(
       exec {"install_plugin_${name}":
         command => $install_cmd,
         creates => "${elasticsearch::plugindir}/${plugin_dir}",
-        onlyif  => "curl -s -XGET 'http://${::ipaddress}:${es_port}/_cluster/health?pretty=true' | awk -F: '/status/{gsub(\"\\\"\",\"\");gsub(\",\",\"\");gsub(\" \",\"\");print \$2}' | grep -q -w green",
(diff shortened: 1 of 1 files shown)


electrical:
Hi,

if there are 2 puppet runs at the same time this won't solve it since the initial check will still be giving a green status.
Perhaps we shouldn't restart ES by default but leave it to the user to restart ES after its all installed?

I've also got https://github.com/elastic/puppet-elasticsearch/issues/363 which draws out a plan i have to make the different places for restarts more controllable.
Would love to have your input.

parabolic:
HI, @electrical , 
yes you are right when puppet is set to run at the same time everywhere my solution doesn't help. https://github.com/elastic/puppet-elasticsearch/issues/363 looks good and I would definitely love to have that, but maybe my option could be present as well and should be turned off by default and with setting a 
{code}
check_cluster_state = true,
{code}
would use the cluster check with onlyif. What do you think ?

electrical:
Hi,

After thinking further about this, i've decided not to merge this since its to specific.
What you could do is set `restart_on_change` to `false` and implement a notify to an exec resource to do the 'green' check and then execute a restart.

PR #391, opened 2015-06-24
make plugin_dir() handle short plugin names properly

Xylakant:
if a short plugin name such as "marvel" is passed instead of a full plugin name, the first element
of the name is returned. This allows calling plugin_dir() on the short plugin name as well, so that
we can stick with the same naming conventions even when passing an url

as discussed on IRC.


electrical:
Hi @Xylakant I've been adding some more fixes lately which should solve all the situations.
Could you verify that the master branch works as expected for you if you have the time?

Thanks!

PR #392, opened 2015-06-24
Re-instate the module_dir option

electrical:


electrical:
@Xylakant could you try out this patch to see if it fixes the issue?
It should re-instate the module_dir option again like it used to work, when not set we will do the auto discovery thing.

PR #405, opened 2015-07-15
Use full-length APT key fingerprint.

jarpy:
Prevents warning from Puppet:

  `The id should be a full fingerprint (40 characters)`


electrical:
Hi,

This is a change i did a while ago but had to revert. See https://github.com/elastic/puppet-elasticsearch/pull/320 for more details.

jarpy:
Thanks. Sorry I didn't spot #320.

PR #406, opened 2015-07-21
Only set owner/group to elasticsearch if needed

sathieu:
Otherwise, binaries and plugins can be hijacked if arbitrary code is executed.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -26,11 +26,6 @@ class elasticsearch::config {
 
   #### Configuration
 
-  File {
-    owner => $elasticsearch::elasticsearch_user,
-    group => $elasticsearch::elasticsearch_group,
-  }
-
   Exec {
     path => [ '/bin', '/usr/bin', '/usr/local/bin' ],
     cwd  => '/',
@@ -50,7 +45,7 @@ class elasticsearch::config {
 
     file { $elasticsearch::logdir:
       ensure  => 'directory',
-      group   => undef,
+      owner   => $elasticsearch::elasticsearch_user,
       mode    => '0644',
       recurse => true,
     }
@@ -60,7 +55,9 @@ class elasticsearch::config {
     }
 
     file { $elasticsearch::datadir:
-      ensure  => 'directory',
+      ensure => 'directory',
+      owner  => $elasticsearch::elasticsearch_user,
+      group  => $elasticsearch::elasticsearch_group,
     }
 
     file { "${elasticsearch::homedir}/lib":
@@ -71,7 +68,7 @@ class elasticsearch::config {
     if $elasticsearch::params::pid_dir {
       file { $elasticsearch::params::pid_dir:
         ensure  => 'directory',
-        group   => undef,
+        owner   => $elasticsearch::elasticsearch_user,
         recurse => true,
       }
 
--- a/spec/acceptance/020_usergroup_spec.rb
+++ b/spec/acceptance/020_usergroup_spec.rb
@@ -50,7 +50,6 @@ describe "elasticsearch class:" do
 
     describe file('/usr/share/elasticsearch') do
       it { should be_directory }
-      it { should be_owned_by 'esuser' }
     end
 
     describe file('/var/log/elasticsearch') do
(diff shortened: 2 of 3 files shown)


sathieu:
This patch is in public domain, no CLA needed.

electrical:
Hi,

Regarding the change, There are more places where i do the same with very good reasons.
I don't exactly see what the issue is with the method I'm using.
Could you provide more context?

Thanks.

sathieu:
@electrical Good security practice is to ensure that a daemon couldn't write it's own bins and libs. This avoid a complete family of potential remote execution.

Also, you use recurse in several places. This is not efficient as Puppet needs to traverse all directories. Permissions should be good from the package manager.

electrical:
Hi,

The recursive is done because one can run the process under a different user if required which is not handled by the package manager.
Also i still don't see what your change improves except setting the user/group parts per resource instead of globally for all file resources.

sathieu:
The patch avoid setting bin and lib dir user and group. If you really want to set them, set them to root:root (for security reasons).

electrical:
The plugin binary is executed under the user the module is set to run as, because of that the binary has to have the right owner/group. See https://github.com/elastic/puppet-elasticsearch/blob/master/manifests/plugin.pp#L85

sathieu:
Why should a program be owner by the executor?? Take a look at /usr/bin ;-)

bflad:
Currently the global user overwrite does not allow our deployment system to automatically update files in the scripts directory.

sathieu:
@bflad I'm not sure to understand. Does this PR help your case?

stevenpost:
Hi,

I'm trying out this module as well.
I concur with sathieu and would even go a step further.
- A process should not be able to change its own binaries or libraries
- A process should not be able to change its own configuration, although many do and refuse to work otherwise (I'm looking at you JBoss)
- Don't needlessly change the ownership and permissions of files, this is especially cumbersome with the plugins (I disabled the recurse there in my own repository, sathieu hasn't mentioned that one yet)

The access permission from my point of view would be:
root:root 0644 for most files
root:root 0755 for executables and directories
esuser:esgroup 0655 for log files.

This can in some cases be tightened when we speak about certificates and passwords, but I don't know if that applies here.

Regards,
Steven

electrical:
Hi,

I understand this is a sensitive topic and that security is obviously important.
Before reaching version 1.0 i would like to tackle this and get this closed down as much as possible while ensuring it all keeps on working.

If we can compile an explicit list of things that need to be changed I'm sure we can get to an agreement and start working on the implementation.

stevenpost:
I think this PR is definitely a step in the right direction, although in my opinion the plugin directory should be in there as well. Although that is a little less straight-forward to do, since I'm not familiar with ES itself and the plugins.

*If* I have some time I'll try to get a PR based on sathieu's work for all these changes, but no promises.

As a side effect, it may become possible to run multiple instances with different users. While we definitely won't have anything like that in production, it may be something useful for a development or test environment. Although this may be a bit out of scope here.

electrical:
This PR only catches a little part of all the user stuff since there are more files that control that ( like `instance.pp` ) I would like to see a PR that handles everything and includes tests as well.

sathieu:
@stevenpost I agree with your rules., but I also think that Puppet should not manage every file (as it is slow, especially using recurse).

This puppet module manages both installation via package (which handle permissions properly) and manual installation (which don't). "recurse" is mostly needed for the later. Currently, recurse is used for:
* $elasticsearch::params::logdir
* "${elasticsearch::params::homedir}/bin"
* $elasticsearch::plugindir
* "${elasticsearch::homedir}/lib"
* $elasticsearch::params::pid_dir
* $instance_configdir

Some of those may have a lot of files (logdir, plugindir). 

Also note that I currently don't have time to handle this PR. Anyone?

sathieu:
Hello, I've updated the patch. Can it be merged, as a step 1 (I agree that there are remaining things, but at least this is better).

tylerjl:
Hey @sathieu, thanks for picking this back up. It looks like some of the tests will need to get updated to reflect the ownership changes (the travis tests should have relevant summaries listing out the specific tests that need updating).

fatmcgav:
Hi there,

Thanks again.

sathieu:
Abandoning this PR. I'm not using puppet anymore.

PR #408, opened 2015-07-22
Export JAVA_HOME in the Redhat init script.

philipwigg:
Fix a bug in the Redhat init script that's already been fixed upstream.

https://github.com/elastic/elasticsearch/issues/5433


electrical:
Oh wow, i really need to stay more on top of init script changes. Will merge this for now but most likely will sync init scripts again soon.

Thanks!

PR #410, opened 2015-07-23
Add offline plugin installation

electrical:
This allowes for the plugin to be served from local filesystem or
puppetmaster.


electrical:
jenkins, test it

PR #427, opened 2015-08-17
Use absolute class names incase someone creates a different Hash/Arra…

benpillet:
…y class in a different namespace

### DIFF
--- a/templates/etc/elasticsearch/elasticsearch.yml.erb
+++ b/templates/etc/elasticsearch/elasticsearch.yml.erb
@@ -33,7 +33,7 @@
         return_vals << stuff
       end
     else
-      if val.is_a?(Hash)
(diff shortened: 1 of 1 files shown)


electrical:
Hi,

is this something you ran into?
I thought it was fairly isolated?

benpillet:
Yep, I ran into this on my setup because of a legacy code issue where is_a? Hash ends up failing.  I think it's because of one of Datadog's templates at https://github.com/DataDog/puppet-datadog-agent/blob/master/templates/agent-conf.d/nginx.yaml.erb

electrical:
Ahh okay. Never expected this :-)
I'll let Jenkins do its magic and merge it when its green.
Thank you for the contribution!

electrical:
jenkins, test it

PR #432, opened 2015-08-21
fix regex retrieving proxy port while downloading plugin

Akanoa:
issue: if url is 'http://my_proxy.domain.tld:3128 your regexp will return "312" instead of "3128"

### DIFF
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -108,7 +108,7 @@ define elasticsearch::plugin(
   }
   elsif ($elasticsearch::proxy_url != undef) {
     $proxy_host_from_url = regsubst($elasticsearch::proxy_url, '(http|https)://([^:]+)(|:\d+).+', '\2')
-    $proxy_port_from_url = regsubst($elasticsearch::proxy_url, '(http|https)://([^:]+)?(:(\d+)).+', '\4')
+    $proxy_port_from_url = regsubst($elasticsearch::proxy_url, '(?:http|https)://[^:/]+(?::([0-9]+))?(?:/.*)?', '\1')
     
     # validate parsed values before using them
     if (is_string($proxy_host_from_url) and is_integer($proxy_port_from_url)) {


electrical:
@Akanoa thank you for the PR.
I'm quite surprised you ran into this issue since the unit test https://github.com/elastic/puppet-elasticsearch/blob/master/spec/defines/004_elasticsearch_plugin_spec.rb#L91-L104 passes all the time.
Is this because my test has a '/' at the end?

electrical:
@Akanoa I've tried your modification but it fails now when the proxy url has a '/' at the end.
Proxy host: localhost
Proxy port: 8080/

Could you ensure it works for both?

Akanoa:
Ok, I see to work your regex, must have a proxy url finished by "/".
I didn't know that it was mandatory.

electrical:
@Akanoa I personally would like to see it work with and without the trailing '/' :-)

Akanoa:
Strange,

That script http://pastebin.com/rQg4yyAF

Produces this stdout :(
puppet apply regexp.pp 
    Notice: Compiled catalog for server.org in environment production in 0.01 seconds
    Notice: regexp [ http://my_proxy.domain.tld:3128 ] > 312
    Notice: /Stage[main]/Main/Notify[regexp [ http://my_proxy.domain.tld:3128 ] > 312]/message: defined 'message' as 'regexp [ http://my_proxy.domain.tld:3128 ] > 312'
    Notice: regexp [ http://my_proxy.domain.tld:3128/ ] > 3128
    Notice: /Stage[main]/Main/Notify[regexp [ http://my_proxy.domain.tld:3128/ ] > 3128]/message: defined 'message' as 'regexp [ http://my_proxy.domain.tld:3128/ ] > 3128'
    Notice: Finished catalog run in 0.18 seconds

electrical:
Sorry i wasn't clear. It currently does that indeed. I would to have it fixed so both options work.

Akanoa:
My bad, I've read and replied too quick :)

Akanoa:
@electrical  I've made a fix about this annoying trailing "/" :)

electrical:
Cool! I'll do some local tests here before merging.
Are you able to sign the CLA as well so i can merge it? thanks!

Akanoa:
I've tried to sign CLA twice but it doesn't work, whereas I've received the confirmation mail from adobe...

electrical:
Ah okay. thanks. I'll have a look at our systems.

electrical:
@Akanoa it seems the Commit email address is different then your github email address. Can you ensure they are the same? That's why its failing atm.

Did some tests and it passes correctly, thank you!

Cheers.

Akanoa:
Ok I'll repush the branch with correct email

Akanoa:
@electrical done

electrical:
Cool, i see the second commit is correct but the first commit has the wrong one.
Sorry to be such a pain, could you fix that one as well? Thank you.

Akanoa:
@electrical done

electrical:
Thank you very much for the contribution!

Akanoa:
You're welcome! :D

PR #434, opened 2015-08-21
Fix small bug causing type error

electrical:
In a very specific situation the plugin define could cause a type error.


electrical:
jenkins, test it

PR #436, opened 2015-08-25
Require the config before starting the service

stevenpost:
Hi,

This PR fixes an issue only seen with restart_on_change => true.
Due to the missing requires, Puppet sometimes tries to start the service before setting the config of the instance, resulting in a failure. This isn't seen in the default configuration because the notify wiil force the service to be delayed till after the (re)start.

Note that I am unable to run the test suite as the following command returns an error:
$ bundle install --path vendor/bundle
...
Gem::InstallError: i18n requires Ruby version >= 1.9.3.
An error occurred while installing i18n (0.7.0), and Bundler cannot continue.
Make sure that `gem install i18n -v '0.7.0'` succeeds before bundling.

Regards,
Steven

### DIFF
--- a/manifests/instance.pp
+++ b/manifests/instance.pp
@@ -279,6 +279,7 @@ define elasticsearch::instance(
       mode    => '0644',
       notify  => $notify_service,
       require => Class['elasticsearch::package'],
+      before  => Elasticsearch::Service[$name],
     }
 
     $require_service = Class['elasticsearch::package']
--- a/spec/defines/005_elasticsearch_instance_spec.rb
+++ b/spec/defines/005_elasticsearch_instance_spec.rb
@@ -206,6 +206,8 @@ describe 'elasticsearch::instance', :type => 'define' do
         let(:pre_condition) { 'class {"elasticsearch": }'  }
 
           it { should contain_elasticsearch__service('es-01').with(:init_template => "elasticsearch/etc/init.d/elasticsearch.#{initscript}.erb", :init_defaults => {"CONF_DIR"=>"/etc/elasticsearch/es-01", "CONF_FILE"=>"/etc/elasticsearch/es-01/elasticsearch.yml", "LOG_DIR"=>"/var/log/elasticsearch/es-01", "ES_HOME"=>"/usr/share/elasticsearch"}) }
+          it { should contain_file('/etc/elasticsearch/es-01/elasticsearch.yml').that_comes_before('Elasticsearch::Service[es-01]') }
+          it { should contain_file('/etc/elasticsearch/es-01/logging.yml').that_comes_before('Elasticsearch::Service[es-01]') }
 
       end
 


stevenpost:
Update: I refuse to sign any CLA (or even read it).
I hereby declare this patch public domain and/or following the license of the existing module.

electrical:
@stevenpost do you mean this happens with `restart_on_change` set to false? since it defaults to true.

stevenpost:
Yes, exactly, sorry I mixed the two around in my initial comment.

electrical:
Been trying to reproduce this issue for a while but without any luck.
Could you share your puppet parts that could reproduce it?
Thank you.

stevenpost:
I'll paste the relevant code below, but you might get a better proof by studying the generated graphs when running with puppet 'agent --test --graph'.
You can use 'zgrviewer' [1] which has a search function (works on multiple OSes, written in Java).
The resources are executed in a mostly random order when running Puppet, so you might not notice it at first. I noticed it because the service tried to start (and failed) before the config was written.
You'll see in the 'relationships.dot' that the resource 'File[/etc/elasticsearch/elkcl2/elasticsearch.yml]' has dependencies (parent directories, the Class[Elasticsearch::Package] resource, its user and group) but no dependants whatsoever (like the service).

[1] http://zvtm.sourceforge.net/zgrviewer.html

```
  class { '::profile::elasticsearch':
    cluster_name    => 'elkcl2',
    heap_size       => '8G',
    instance_config => {
      'bootstrap.mlockall'                   => 'true',
      'cluster.name'                         => 'elkcl2',
      'node.max_local_storage_nodes'         => '1',
      'index.number_of_replicas'             => '0',
      'index.number_of_shards'               => '1',
      'network.publish_host'                 => $::fqdn,
      'discovery.zen.ping.multicast.enabled' => 'false',
      'discovery.zen.ping.unicast.hosts'     => "[\"${::fqdn}\"]",
      'action'                               => {
        'disable_delete_all_indices' => 'true',
        'destructive_requires_name'  => 'true',
        'disable_shutdown'           => 'true',
      },
      'script.disable_dynamic'               => 'false',
    },
  }

class profile::elasticsearch (
  $cluster_name,
  $heap_size,
  $instance_config,
) {

  # create the elasticsearch group with a specific GID.
  group { 'elasticsearch':
    ensure => present,
    gid    => 1499,
  }

  user { 'elasticsearch':
    ensure     => present,
    uid        => 1499,
    gid        => 1499,
    shell      => '/bin/bash',
    home       => '/usr/share/elasticsearch',
    managehome => true,
    require    => Group['elasticsearch'],
  }

  file { '/usr/share/elasticsearch/.profile':
    ensure  => 'present',
    mode    => '0664',
    owner   => 'elasticsearch',
    group   => 'elasticsearch',
    source  => 'puppet:///modules/profile/elasticsearch/profile',
    require => User['elasticsearch'],
  }

  $_es_dirs = ['/var/app/elasticsearch', '/var/app/elasticsearch/logs', '/var/app/elasticsearch/bin']
  file { $_es_dirs :
    ensure => 'directory',
    owner  => 'elasticsearch',
    group  => 'elasticsearch',
    mode   => '0664',
  }
  ->
  # version lock via hiera
  class { '::elasticsearch':
    datadir           => '/var/app/elasticsearch/data',
    restart_on_change => false,
    require           => User['elasticsearch'],
  }

  elasticsearch::instance { $cluster_name :
    config        => $instance_config,
    init_defaults => {
      'ES_HEAP_SIZE'   => $heap_size,
      'MAX_OPEN_FILES' => '65535',
      'MAX_MAP_COUNT'  => '262144',
      'LOG_DIR'        => "/var/app/elasticsearch/logs/${cluster_name}",
    },
  }

  file { '/var/app/elasticsearch/bin/elasticsearch-remove-old-indices.sh':
    content => template('roles/elk/elasticsearch-remove-old-indices.sh.erb'),
    owner   => 'elasticsearch',
    group   => 'elasticsearch',
    mode    => '0744',
  }

  elasticsearch::plugin { 'elasticsearch/marvel/1.3.1':
    instances => $cluster_name,
  }
  elasticsearch::plugin { 'mobz/elasticsearch-head':
    instances => $cluster_name,
  }
  elasticsearch::plugin { 'lmenezes/elasticsearch-kopf':
    instances => $cluster_name,
  }
  elasticsearch::plugin { 'lukas-vlcek/bigdesk':
    instances => $cluster_name,
  }
  elasticsearch::plugin { 'royrusso/elasticsearch-HQ':
    instances => $cluster_name,
  }

}
```

electrical:
Been checking it out further.
With `restart_on_change` set to false its correct it will never trigger the services dependency since that would trigger the restart ( which with restart_on_change is set not to do that )
So your conclusion is correct.

The config and logging files for the instance are still dependent on the right directories and stuff so that looks good as well.
Because of this i don't see any direct issues.

I also did a test of setting up an instance ( fresh ) directly with `restart_on_change` set to t false which seems to work fine since it will boot up the new instance but any subsequent changes do not trigger into a restart.

stevenpost:
Have you checked the graphs?
It is correct that the 'notify' shouldn't be there, but the 'require' is missing.
A 'require' doesn't trigger a service refresh (restart), just ensures the ordering, while a 'notify' does both.
There is no harm in specifying them both.

You could force the issue to occur by defining a third resource that depends on the service, and define a relationship to force the config after that resource:
Service['es_service'] -> File['/etc/thirdresource'] -> File['config_file']
That line, as well as the third resource can be placed outside the Elastic search module.
What you will see is that Puppet honors this dependency, whereas if the ordering was correct (enforced in any way) you would get a compile error because of a dependency cycle.
On second thought, you don't even need that third resource.

The issue here is the initial run, since any subsequent run will finish without problems as the file has been created anyway, and the service will be started in the second run.

electrical:
Been doing some more digging in the instance code.
It seems that most parts do a before on the service part except for the config file.
I've never seen any issues with it so far but that might solve the issue.
Could you check and verify?

stevenpost:
A 'before' is the exact opposite of a 'require' in Puppet code, so that would certainly solve the issue.
It might even be preferable if that is the current style of the module. Looking into the code again, I see that the logging.yml file already has a 'before', so that part of my patch wouldn't even have any effect.
I only observed the issue with the elasticsearch.yml file (and never after my patch).

I'll update the PR accordingly.

stevenpost:
Could you verify the rspec tests? I'm not entirely sure they will succeed now, and sonce I can't run the tests....
This changed patch also makes it a lot easier for you to see what it does, since it's a single line in the manifests.

electrical:
The unit tests require at least ruby 1.9.3 indeed, you are most likely running 1.8.7 ?
I'll let Jenkins do its magic and test it.

electrical:
jenkins, test it

stevenpost:
Hi,

Can you tell me why it failed?
On first sight I see every test in the log, not just wh

stevenpost:
> The unit tests require at least ruby 1.9.3 indeed, you are most likely running 1.8.7 ?
Yes, our machines all run RHEL 6. Perhaps you can do something similar to the Apache module for Puppet? There I can run the tests if I do a 'bundle install --path=vendor/gems --without-system-tests'.

electrical:
@stevenpost the 'requires' parts that you added are failing on the service.
You'll need to add the 'before' part on the 2 files.

Re ruby version; in my case also the unit tests require 1.9.3 or higher due to a library i use for junit reporting.
I'll have a look to see if i can organize it better.
But you could also use something like rbenv or rvm for ruby version management so you can run newer versions.

stevenpost:
I've changed the unit tests, just hold on while I try to get the tests running with a more recent ruby version. I'll try to run them using RHSCL, but this may hide issues you can see with 1.8.7, as I'm not going to install 1.9.3 on over 500 machines just for this.

I'll report back with the results.

stevenpost:
Update: the complete run took 29 minutes and 47 seconds on this testing VM, but completed without an error.

Using RHSCL (Red Hat Software Collections) works fine, I used ruby 1.9.3 in this case.

electrical:
Cool. thank you. Will trigger Jenkins to do a full test suite as well.

electrical:
jenkins, test it

stevenpost:
Sorry, my bad.
I was testing master instead of my fixed branch... -.- will look into it.

stevenpost:
Fixed for real this time :)

electrical:
@stevenpost could you rebase against the master branch? seems several things have changed along the way.
Also we do need the CLA signed.

stevenpost:
The added datacat seems to complicate matters a bit, I will look into it if/when the CLA issue is resolved.
Please see https://github.com/elastic/puppet-elasticsearch/pull/437#issuecomment-140333011 for that.

electrical:
Datacat was required for other modules to interact with the ES module for config parts.
For the CLA we have started an internal discussion.

tylerjl:
If you do happen to change your mind in regards to the CLA, feel free to re-open and we can revisit. Thank you for your work regardless!

PR #437, opened 2015-08-31
Honor the plugindir setting when installing plugins

stevenpost:
Hi,

This PR allows plugins to be installed in the proper directory.
When changing the plugindir, plugins keep getting installed into the default directory (/usr/share/elasticsearch/plugins)'. This causes a Puppet run to fail since the check for installed plugins *is* changed.

Use case:
The people initially setting up Elasticsearch at our site opted for a separate filesystem, mounted on '/es' with some subdirectories:
/es/bin (helperscripts)
/es/logs
/es/data
/es/plugins

Regards,
Steven

PS: no CLA required, this patch follows the existing license, which is Apache License, Version 2.0.

### DIFF
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -138,10 +138,10 @@ define elasticsearch::plugin(
   }
 
   if ($real_url == undef) {
-    $install_cmd = "${elasticsearch::plugintool}${proxy} install ${name}"
+    $install_cmd = "${elasticsearch::plugintool}${proxy} install -Des.path.plugins=${elasticsearch::plugindir} ${name}"
     $exec_rets = [0,]
   } else {
-    $install_cmd = "${elasticsearch::plugintool}${proxy} install ${name} --url ${real_url}"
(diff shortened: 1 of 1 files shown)


electrical:
Hi,

I'm afraid that the plugindir option is not fully functional since it doesn't set the right path in the elasticsearch.yml file. That's something i still need to fix.
That said, I'm afraid we do need you to sign the CLA for any of the contributions.
You will only require to do this once and this will count for any of the elastic projects.

stevenpost:
Hi,

I said it once, I'll say it again now, but that is the last time:
I will not sign anything. You will either accept the patches I write under the license of the project (Apache License 2) or not incorporate them. This is not up for debate.

Please read the following and consider abandoning the CLA:
http://ebb.org/bkuhn/blog/2014/06/09/do-not-need-cla.html

Another interesting read on the subject is this:
http://opensource.com/law/11/7/trouble-harmony-part-1

You may need to get in touch with your legal department on this.

Regards,
Steven

PR #443, opened 2015-09-02
Shorten spec run by only looping over specific parts

electrical:
Finished in 10 minutes 31 seconds (files took 6.14 seconds to load)
1480 examples, 0 failures

vs

Finished in 20.97 seconds (files took 5.81 seconds to load)
93 examples, 0 failures

### DIFF
--- a/spec/defines/005_elasticsearch_instance_spec.rb
+++ b/spec/defines/005_elasticsearch_instance_spec.rb
@@ -55,324 +55,340 @@ describe 'elasticsearch::instance', :type => 'define' do
       let(:title) { 'es-01' }
       let(:pre_condition) { 'class {"elasticsearch": }'  }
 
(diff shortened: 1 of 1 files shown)


spec/defines/005_elasticsearch_instance_spec.rb
```diff
@@ -55,324 +55,340 @@
...
-        context "set a value" do
+  let :facts do {
+    :operatingsystem => 'CentOS',
+    :kernel => 'Linux',
+    :osfamily => 'RedHat',
+    :operatingsystemmajrelease => '6',
+    :scenario => '',
+    :common => '',
+    :hostname => 'foo'
+  } end
```
igalic: you still use the facts gem, but just use .first, instead of .each ;)

PR #450, opened 2015-09-13
Update SHA for compatibility with latest version of apt module

hashbrowncipher:
The puppetlabs-apt module now emits a warning when short fingerprints are used.
This quiets that warning.


hashbrowncipher:
Closing due to #347

PR #452, opened 2015-09-14
Clean up apt block

pcfens:
Remove the conditional apt block (replacing it with simple `include ::apt`). Also explicitly state that `apt::update` should come before any packages are installed.

### DIFF
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -39,7 +39,7 @@ class elasticsearch::repo {
         location    => "http://packages.elastic.co/elasticsearch/${elasticsearch::repo_version}/debian",
         release     => 'stable',
         repos       => 'main',
-        key         => '46095ACC8548582C1A2699A9D27D666CD88E42B4',
+        key         => 'D88E42B4',
         key_source  => 'http://packages.elastic.co/GPG-KEY-elasticsearch',
         include_src => false,
       }
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -32,15 +32,14 @@ class elasticsearch::repo {
 
   case $::osfamily {
     'Debian': {
-      if !defined(Class['apt']) {
-        class { 'apt': }
-      }
+      include ::apt
+      Class['apt::update'] -> Package[$elasticsearch::package_name]
 
       apt::source { 'elasticsearch':
         location    => "http://packages.elastic.co/elasticsearch/${elasticsearch::repo_version}/debian",
         release     => 'stable',
         repos       => 'main',
-        key         => 'D88E42B4',
+        key         => '46095ACC8548582C1A2699A9D27D666CD88E42B4',
         key_source  => 'http://packages.elastic.co/GPG-KEY-elasticsearch',
         include_src => false,
       }


manifests/repo.pp
```diff
@@ -32,15 +32,14 @@
...
-      }
+      include ::apt
+      Class['apt::update'] -> Package<| |>
 
       apt::source { 'elasticsearch':
         location    => "http://packages.elastic.co/elasticsearch/${elasticsearch::repo_version}/debian",
         release     => 'stable',
         repos       => 'main',
-        key         => 'D88E42B4',
+        key         => '46095ACC8548582C1A2699A9D27D666CD88E42B4',
```
electrical: the long key doesn't work with older puppetlabs-apt module versions which are used by older PE versions.
Once i'm able to drop those older versions i can use the longer key.

manifests/repo.pp
```diff
@@ -32,15 +32,14 @@
 
   case $::osfamily {
     'Debian': {
-      if !defined(Class['apt']) {
-        class { 'apt': }
-      }
+      include ::apt
+      Class['apt::update'] -> Package<| |>
```
electrical: this should already be handled by the apt module it self, also when defining this relation multiple times with other modules might cause problems ?

manifests/repo.pp
pcfens: Which versions of the apt module need to be supported? I'm not a PE user so I'm not familiar with it, though I do know that 1.8 supports the 40 character key ID.

electrical: 3.3.x doesn't support it yet. I'm also not sure about 3.7.x  I need to do some testing on it which apt module's get installed.

manifests/repo.pp
```diff
@@ -32,15 +32,14 @@
 
   case $::osfamily {
     'Debian': {
-      if !defined(Class['apt']) {
-        class { 'apt': }
-      }
+      include ::apt
+      Class['apt::update'] -> Package<| tag == 'elasticsearch-package' |>
```
pcfens: Even though something like this would ideally be handled by the `apt` module, requiring `apt::update` before packages is a [known limitation](https://github.com/puppetlabs/puppetlabs-apt#limitations). Using tags on the package resource should avoid help avoid dependency cycles since the collector is limited to packages we install.

electrical: Since repo management is automatically done before package install via init.pp does it make sense to do a notify on `Class['apt::update']` after the `apt::source` part?

pcfens: If I'm understanding things correctly, we can't be certain that resources in `apt::update` (in this case the `exec` resource) will be handled before package installs since `apt::update` is included, but not [contained](https://docs.puppetlabs.com/puppet/latest/reference/lang_containment.html), inside of  the `apt` class (containing it there would likely cause other issue). Adding a notify would ensure that `apt::update` is handled after `apt::source`, but wouldn't ensure it all happens prior to the handling the `package` resource.

There is however a cleaner method than a tag that I'll push up shortly.

manifests/repo.pp
electrical: PE 3.3.2 has puppetlabs-apt 1.5.0 included so can't upgrade that and it supports only the short notations.
Unless i drop 3.3.2 support ( which i most likely will do before 1.0 release ) the key will have to be the short version.
If you could change that i can merge this and hopefully solve the apt 2.x issue :-)

pcfens:
Reverted to short key and rebased

electrical:
jenkins, test it

electrical:
jenkins, test it

PR #465, opened 2015-10-19
Use apt 40 char length key

jpoittevin:
Since apt module displays a warning if not


electrical:
Hi,

Due to PE 3.7.x support I'm unable to do this yet since it ships with an older version of pl-apt.

PR #476, opened 2015-11-02
Fix plugin remove for 2.x

cyberhuman:
bin/plugin does not have "--remove" option now, "remove" must be used instead.


electrical:
Hi,

Thank you for the PR.
Since this module still needs to work with ES 1.x releases I will have to check if this change works there as well.
Since there are some more changes for ES 2.x required that are not backwards compatible i will most likely have to do 2 releases of the module and have 2 branches so that i can still work on ES 1.x things as well.

electrical:
Hi,

Thank you very much for the PR but I've decided to merge #477 since it allows for backwards compat working with ES 1.x
PR #483 adds the test as well for ES 2.x
I'm planning to do a release early this week.

cyberhuman:
Hi. I'm ok with that, thank you!

PR #477, opened 2015-11-02
allow installation of plugins on 2.x

igalic:
this adds a fact, elasitcsearch_version, which works without
elasticsearch running, and helps determine the basic version of the
es installation.

using this info, we then decide in elasticsearch::plugin how to
(un)install plugins.

### DIFF
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -77,7 +77,7 @@ define elasticsearch::plugin(
     $proxy_port  = undef,
 ) {
 
-  include elasticsearch
+  include ::elasticsearch
 
   Exec {
     path      => [ '/bin', '/usr/bin', '/usr/local/bin' ],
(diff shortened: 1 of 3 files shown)


igalic:
elasticsearch::plugin should probably be elasticsearch_plugin…

electrical:
@igalic thanks!
Was looking for a good solution on this but completely forgot the cmd line version option :-)
Will trigger jenkins on this.

electrical:
jenkins, test it

electrical:
jenkins, test it

electrical:
Did some local checking.
`--version` only works on 2.x
on 1.x its `-v`

I hope this is taken in account? if so could you also add some tests for this? ( Unit tests )

manifests/plugin.pp
```diff
@@ -93,6 +93,11 @@
     default => Elasticsearch::Service[$instances],
   }
 
+  $remove_cmd = $::elasticsearch_version? {
+    /^2/    => 'remove',
```
igalic: this should use `versioncmp()`

electrical:
Could you add a test to the plugin define as well?

igalic:
yes!

…

how?

electrical:
I think we can set the fact value and just assert we set the right install / remove commands? Worst case i can add them afterwards :-)

PR #486, opened 2015-11-09
repo.pp update key to full fingerprint

cotocisternas:
Apt_Key Warning: 'The id should be a full fingerprint (40 characters), see README.'


electrical:
Hi. Due to older PE versions not supporting this yet, i'm unable to merge the long key.
As soon those are dropped this is possible.

cotocisternas:
What about if i extract the fingerprint to a param, and defaulted to the short one?

electrical:
Hi,

I replied to the PR you sent for that. I'm +1 on that once Jenkins is green. Thanks!

PR #491, opened 2015-11-13
Some changes to support OpenBSD:

buzzdeee:
* a couple of OpenBSD specific additions to params.pp, i.e. user,
   download tool, paths, service provider, etc.
 * add a $service_flags parameter to instance.pp, since the OpenBSD
   service type supports flags, and those should be settable.
 * add a $service_flags parameter to service.pp, that gets passed
   in from instance.pp, and handed over to openbsd.pp service.
 * add an easy openbsd service provider, that only knows about
   status, ensure, pid_dir, service_flags and init_template
   parameters. There is no "standard" defaults file on OpenBSD.
 * because the defaults_location on OpenBSD is not set, guard
   the augeas call in config.pp to manage the defaults file.
 * added a template for the openbsd service, based on the
   rc script, that comes with the package.

tested with elasticsearch-1.7.2, configured and fired up an instance as well as plugin installation works.

Let me know if there is something that needs change/enhancement to get merged.

thanks,
Sebastian

### DIFF
--- a/lib/puppet/provider/elasticsearch_plugin/plugin.rb
+++ b/lib/puppet/provider/elasticsearch_plugin/plugin.rb
@@ -4,8 +4,15 @@ Puppet::Type.type(:elasticsearch_plugin).provide(:plugin) do
   desc "A provider for the resource type `elasticsearch_plugin`,
         which handles plugin installation"
 
-  commands :plugin => '/usr/share/elasticsearch/bin/plugin'
-  commands :es => '/usr/share/elasticsearch/bin/elasticsearch'
+  os = Facter['osfamily'].value
+  if os == 'OpenBSD'
+    commands :plugin => '/usr/local/elasticsearch/bin/plugin'
+    commands :es => '/usr/local/elasticsearch/bin/elasticsearch'
+    commands :javapathhelper => '/usr/local/bin/javaPathHelper'
+  else
+    commands :plugin => '/usr/share/elasticsearch/bin/plugin'
+    commands :es => '/usr/share/elasticsearch/bin/elasticsearch'
+  end
 
   def exists?
     es_version
@@ -100,11 +107,23 @@ Puppet::Type.type(:elasticsearch_plugin).provide(:plugin) do
 
   def es_version
     return @es_version if @es_version
+    es_save = ENV['ES_INCLUDE']
+    java_save = ENV['JAVA_HOME']
+
+    os = Facter['osfamily'].value
+    if os == 'OpenBSD'
+      ENV['JAVA_HOME'] = javapathhelper('-h', 'elasticsearch').chomp
+      ENV['ES_INCLUDE'] = '/etc/elasticsearch/elasticsearch.in.sh'
+    end
     begin
       version = es('-version')
     rescue
+      ENV['ES_INCLUDE'] = es_save if es_save
+      ENV['JAVA_HOME'] = java_save if java_save
       raise "Unknown ES version. Got #{version.inspect}"
     ensure
+      ENV['ES_INCLUDE'] = es_save if es_save
+      ENV['JAVA_HOME'] = java_save if java_save
       @es_version = version.scan(/\d+\.\d+\.\d+(?:\-\S+)?/).first
       debug "Found ES version #{@es_version}"
     end
--- a/manifests/instance.pp
+++ b/manifests/instance.pp
@@ -64,6 +64,9 @@
 # [*init_defaults_file*]
 #   Defaults file as puppet resource
 #
+# [*service_flags*]
+#   Service flags used for the OpenBSD service configuration, defaults to undef.
+#
 # === Authors
 #
 # * Richard Pijnenburg <mailto:richard.pijnenburg@elasticsearch.com>
@@ -78,6 +81,7 @@ define elasticsearch::instance(
   $logging_config     = undef,
   $logging_template   = undef,
   $logging_level      = $elasticsearch::default_logging_level,
+  $service_flags      = undef,
   $init_defaults      = undef,
   $init_defaults_file = undef,
   $init_template      = $elasticsearch::init_template
@@ -307,6 +311,7 @@ define elasticsearch::instance(
   elasticsearch::service { $name:
     ensure             => $ensure,
     status             => $status,
+    service_flags      => $service_flags,
     init_defaults      => $init_defaults_new,
     init_defaults_file => $init_defaults_file,
     init_template      => $init_template,
--- a/manifests/package.pp
+++ b/manifests/package.pp
@@ -38,21 +38,21 @@ class elasticsearch::package {
   if $elasticsearch::ensure == 'present' {
 
     # Create directory to place the package file
+    $package_dir = $elasticsearch::package_dir
     exec { 'create_package_dir_elasticsearch':
       cwd     => '/',
       path    => ['/usr/bin', '/bin'],
-      command => "mkdir -p ${elasticsearch::package_dir}",
-      creates => $elasticsearch::package_dir,
+      command => "mkdir -p ${package_dir}",
+      creates => $package_dir,
     }
 
-    file { $elasticsearch::package_dir:
+    file { $package_dir:
       ensure  => 'directory',
       purge   => $elasticsearch::purge_package_dir,
       force   => $elasticsearch::purge_package_dir,
       backup  => false,
       require => Exec['create_package_dir_elasticsearch'],
     }
-
 
     # Check if we want to install a specific version or not
     if $elasticsearch::version == false {
@@ -77,7 +77,6 @@ class elasticsearch::package {
         default:   { fail("software provider \"${elasticsearch::package_provider}\".") }
       }
 
-      $package_dir = $elasticsearch::package_dir
 
       $filenameArray = split($elasticsearch::package_url, '/')
       $basefilename = $filenameArray[-1]
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -131,6 +131,7 @@ define elasticsearch::plugin(
         source     => $file_source,
         url        => $url,
         proxy_args => $proxy,
+        plugin_dir => $::elasticsearch::plugindir,
         notify     => $notify_service,
       }
 
--- a/templates/etc/init.d/elasticsearch.OpenBSD.erb
+++ b/templates/etc/init.d/elasticsearch.OpenBSD.erb
@@ -1,0 +1,27 @@
+#!/bin/sh
+#
+# This file is managed via PUPPET
+
+daemon="/usr/local/elasticsearch/bin/elasticsearch"
+daemon_flags="-d -Des.default.path.conf=/etc/elasticsearch/<%= @name %> -p <%= @pid_dir %>/elasticsearch-<%= @name %>.pid"
+daemon_user="_elasticsearch"
+
+. /etc/rc.d/rc.subr
+
+pexp="$(/usr/local/bin/javaPathHelper -c elasticsearch) .*org.elasticsearch.bootstrap.Elasticsearch.*"
+
+rc_reload=NO
+
+rc_start() {
+	${rcexec} \
+		"ES_INCLUDE=\"/etc/elasticsearch/elasticsearch.in.sh\" \
+		"CONF_DIR=\"/etc/elasticsearch\"" \
+		JAVA_HOME=\"$(/usr/local/bin/javaPathHelper -h elasticsearch)\" \
+		${daemon} ${daemon_flags}"
+}
+
+rc_pre() {
+	install -d -o _elasticsearch /var/run/elasticsearch/
+}
+
+rc_cmd $1
--- a/README.md
+++ b/README.md
@@ -127,7 +127,6 @@ elasticsearch::plugin{ 'jetty':
 }
 ```
 
-
 ####Using a proxy
 You can also use a proxy if required by setting the `proxy_host` and `proxy_port` options:
 ```puppet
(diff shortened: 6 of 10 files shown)


electrical:
Hi,

Thank you very much for adding the support! I'll have a good look at this very soon :-)

manifests/instance.pp
```diff
@@ -306,6 +310,7 @@
   elasticsearch::service { $name:
     ensure             => $ensure,
     status             => $status,
+    service_flags      => $init_flags,
```
electrical: You have `service_flags` earlier but here its called `init_flags` ?

templates/etc/init.d/elasticsearch.OpenBSD.erb
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+#
+# $OpenBSD: elasticsearch.rc,v 1.2 2015/07/15 23:34:49 ajacoutot Exp $
+
+daemon="/usr/local/elasticsearch/bin/elasticsearch"
+daemon_flags="-d -Des.config=/etc/elasticsearch/<%= @name %>/elasticsearch.yml -p <%= @pid_dir %>/elasticsearch-<%= @name %>.pid"
```
electrical: I think i'm missing a few options here.
from the redhat init script:

```
-Des.default.path.home=$ES_HOME 
-Des.default.path.logs=$LOG_DIR 
-Des.default.path.data=$DATA_DIR 
-Des.default.path.work=$WORK_DIR 
-Des.default.path.conf=$CONF_DIR
```

Could you perhaps include the same options?

electrical:
Left a few comments.
Could you also test it with ES 2.0 ?

Thanks!

manifests/instance.pp
buzzdeee: indeed, I initially had this as $init_flags, because of all the other init_* variables, but then, since the others don't really make sense that much, I changed it to service_flags, and seem to have this one forgotten, fixed now.

templates/etc/init.d/elasticsearch.OpenBSD.erb
buzzdeee: I only have the few parameters in the file for the daemon_flags, since this is what the "original" rc script from the package sets. Therefore that is all 
Further, there is no concept of /etc/{sysconfig|defaults}/progname or the like like on many (or most) Linux distributions. 
That's why I protected the augeas call in config.pp "${elasticsearch::params::defaults_location}/elasticsearch" to only get evaluated, when that defaults_location parameter is set, which it isn't for OpenBSD.
Also I ripped all the defaults_file related stuff out of service/openbsd.pp.
However, the $service_flags parameter of each instance, the user of the module can override the daemon_flags of each rc script.

i.e. 
service { 'elasticsearch_es01':
  flags => '-d -Des.config=/etc/elasticsearch/es01/elasticsearch.yml -p /var/run/elasticsearch/elasticsearch-es01.pid-Des.default.path.home=/tmp/blubb -Des.default.path.logs=/var/log/mylogs/'
}

that would allow the user to specify whatever they like it to be.
OpenBSD users are usually used to those service flags.

Alternatively, as I did, I configured it for the instances in Hiera yaml:
elasticsearch::instances:
  'es01':
    datadir: '/var/elasticsearch/es-data-es01/'
    logging_level: 'DEBUG'
    config:
      path:
        conf: '/etc/elasticsearch/es01/'
        logs: '/var/log/elasticsearch/'

Both ways should work fine.

buzzdeee:
OpenBSD only currently comes with 1.7.X release of elasticsearch. I'm aware for logstash, a 2.X is in the queue to get added, but I don't know for ES, maybe also soon in the future.
Since I run OpenBSD -current mostly everywhere, whenever ES gets updated to 2.X and I run into trouble with the module, I'll likely open a new PR with patch ;)

cheers,

electrical:
Cool. Thank you for the clarification.
I assume that we don't need to setup any special repositories then ?
I'll kick off jenkins for the current things.
I don't have an openbsd vagrant box to run the tests against but will be doing that at some point so I can be sure not to break things later :)

electrical:
jenkins, test it

buzzdeee:
No extra repo required, it's maintained in the ports tree, and packages available on  all the OpenBSD mirrors.

Note: I use this @home, or just start looking into it, learning. I'm OpenBSD only, and run (more or less current) OpenBSD -current here. Similar aproach to the OpenBSD -current, I run with Puppet modules. I've r10k setup, to fetch all modules I use from its git repo master branch. I don't use them from the forge. When things suddenly start to break for me (I had this in the past with a few modules already) be sure I'll at least complain, or even provide a PR ;)

thanks,
Sebastian

buzzdeee:
BTW, elasticsearch package for OpenBSD was just updated to 2.0 today ;). So will likely test with that new version in a few days.

buzzdeee:
meanwhile tested with ES 2.0.0, and even ES 2.1.0. Basic installation, configuration and service management works as advertized. 
However, plugin management doesn't work with those 2.X version anymore, but I think that is not unique to OpenBSD ;)

electrical:
jenkins, test it

electrical:
@buzzdeee plugin install for 2.x is indeed broken, but working on a PR to fix that.

electrical:
@buzzdeee sorry i haven't looked at this PR anymore. been busy fixing the plugin installation for 2.x which is finally in.
I'll be looking at this PR very soon.

buzzdeee:
@electrical thanks for getting back. I just rebased, and tried again to manage plugins but got odd failures about missing dependencies. I'll look into it in the next few days, when I figure how to fix that, and make the plugins management work, I'll get back here.

electrical:
@buzzdeee sounds good. Feel free to ping me at anytime if you need any help.

buzzdeee:
so, got plugin installation to work on OpenBSD.
Mainly see my commit message, what I did to it. I kept it as a separate commit to ease working on it based on feedback.

I just saw, and maybe that may be preferred:
  * instead of that ugly additional parameter, I could define the plugin_dir like  ${elasticsearch::params::homedir}/plugins, as it is done in scripts.pp for the
scripts directory, Alternatively I could directly use the ${elasticsearch::params::plugindir} 
and hand that over to the native plugin type from plugin.pp.

Then maybe even remove the defaultto of the plugin_dir parameter of the native plugin type?

That should then also give similar/same results on Linux as well as on OpenBSD.

I think proposed approach would make more sense, so let me know if you prefer that, and it should be easy to change for me.

lib/puppet/provider/elasticsearch_plugin/plugin.rb
```diff
@@ -90,7 +98,12 @@ def destroy
   def es_version
     return @es_version if @es_version
     begin
-      version = es('-v') # ES 1.x
+      os = Facter['osfamily'].value
+      if os == 'OpenBSD'
+        version = pkg_info('elasticsearch').lines.first
```
electrical: Don't think this special case is required since getting the version via the elasticsearch binary should work on any platform right?

manifests/plugin.pp
```diff
@@ -75,6 +81,7 @@
     $source      = undef,
     $proxy_host  = undef,
     $proxy_port  = undef,
+    $plugin_dir  = undef,
```
electrical: We can pass the plugin_dir from the params class here so we get the right one depending the distro. This way the user doesn't have to specify it them selves.
Also the plugin dir is set via the main class and passed on internally setting different variables ( including config file )

lib/puppet/provider/elasticsearch_plugin/plugin.rb
buzzdeee: running the plain binary I get:
# /usr/local/elasticsearch/bin/elasticsearch --version
You must set the ES_CLASSPATH var

So I would have to run it like this:
ES_INCLUDE="/etc/elasticsearch/elasticsearch.in.sh" /usr/local/elasticsearch/bin/elasticsearch --version

to get the version output,
so I would have to special case it in any way. 
If you prefer, I can look into setting the environment variable, before attempting to run the command, instead of using pkg_info. Shouldn't be too complicated.

electrical: That's weird it doesn't happen on the debian/ubuntu/centos based machines.
elasticsearch binary should be handling the elasticsearch.in.sh automatically.

electrical: Ah. we don't look in /usr/local/elasticsearch/bin by default at https://github.com/elastic/elasticsearch/blob/master/distribution/src/main/resources/bin/elasticsearch#L79-L95

Would be nice to have that fixed in the actual binary as well.
For the time being you can indeed set the ENV variable before running the command.
At least that way the local stays the same.

manifests/package.pp
```diff
@@ -38,19 +38,22 @@
   if $elasticsearch::ensure == 'present' {
 
     # Create directory to place the package file
-    exec { 'create_package_dir_elasticsearch':
-      cwd     => '/',
-      path    => ['/usr/bin', '/bin'],
-      command => "mkdir -p ${elasticsearch::package_dir}",
-      creates => $elasticsearch::package_dir,
-    }
+    $package_dir = $elasticsearch::package_dir
```
electrical: the package dir is used for puppet:// based plugins so its fairly mandatory to be created, otherwise it will fail.
See https://github.com/elastic/puppet-elasticsearch/blob/master/manifests/plugin.pp#L115

electrical:
Left 1 comment, looks good further!

manifests/package.pp
buzzdeee: see below, its not entirely removed, its only guarded by a if $package_dir.
For OpenBSD, I've set it to undef, see in the params.pp.
Since I used the packages that come with the distribution, I don't have a need for the package_dir variable to be set. However, people should still be able to override it as parameter if really needed.

electrical: In this case the package_dir var is used by the plugin define as well to have a temp location to place the plugin when it comes from a puppet:// resource for offline installations, so i think it needs to be re-implemented for OpenBSD as well.

buzzdeee: ah, that I haven't seen, will look into it, and find a suitable default value.

buzzdeee:
I think I addressed all your points, also got the plugin installation working. Had to export two environment variables, and resetting them afterward, to not pollute the rest of the puppet run with it.
Let me know if there is anything else.

electrical:
@buzzdeee thank you! I'll try to review this today or tomorrow and let jenkins do its magic in the meantime to ensure it doesn't break existing stuff.

electrical:
jenkins, test it

electrical:
Code looks good! Waiting for Jenkins to report back and then i can merge it.
I'll have to look for a vagrant box of OpenBSD to so i can run tests on it to ensure i don't break it in the future.

manifests/plugin.pp
```diff
@@ -77,7 +77,7 @@
     $proxy_port  = undef,
 ) {
 
-  include elasticsearch
+  require elasticsearch
```
electrical: This seems to have caused a dependency cycle.

```
 Error: Could not apply complete catalog: Found 1 dependency cycle:
  (Augeas[defaults_es-01] => Elasticsearch::Service::Init[es-01] => Elasticsearch::Service[es-01] => Elasticsearch::Instance[es-01] => Class[Elasticsearch] => Elasticsearch::Plugin[mobz/elasticsearch-head/1.0.0] => Elasticsearch_plugin[mobz/elasticsearch-head/1.0.0] => Elasticsearch::Plugin[mobz/elasticsearch-head/1.0.0] => Class[Elasticsearch])
```

Don't think we need the require for accessing the variable ( plugindir ) since it always worked with the restart_on_change one.

Also in https://github.com/elastic/puppet-elasticsearch/blob/master/manifests/init.pp#L411 there is already an explicit dependency for it.

buzzdeee: Thanks Mr. Jenkins for spotting that ;)

I tested it for me, and did not ran into recursion issues, therefore kept it.
but I also don't have to deal with that augeas thingie bingie for that defaults file on OpenBSD.

reverted it back to 'include', since the include also seems to be fine for me.

electrical:
jenkins, test it

manifests/plugin.pp
electrical: This hasn't changed yet by the looks of it and is still causing the test failure.

buzzdeee: crap, I changed it, but then squashed/reordered commits and it reappeard without /me noticing.
Now it's back as it should be.

electrical: hehe no worries. can happen.
Will let Jenkins do its thing again. thanks!

electrical:
jenkins, test it

electrical:
Hmm, don't know why jenkins didn't trigger. stupid thing. lets try it again

electrical:
jenkins, test it

electrical:
jenkins, test this

electrical:
jenkins, test it

electrical:
everything looks good, just want to let CI run it a bit more. found some weird testing things.

buzzdeee:
for your convenience, rebased

electrical:
Thank you for the rebase. Made some changes to make testing more stable.
Will let Jenkins do some more work and merge it tomorrow.

electrical:
jenkins, test it

electrical:
jenkins, test it

electrical:
@buzzdeee I'm afraid a rebase is needed. sorry.

buzzdeee:
no worries, rebased yesterday evening, still seems to be fine for me (:

electrical:
jenkins, test it

electrical:
Sorry, this occidentally dropped of my radar last week. Hoping jenkins will do its thing this time and i'll merge it and release next week.

electrical:
jenkins, test it

buzzdeee:
no worries (:

PR #494, opened 2015-11-13
repo_key_id to param

cotocisternas:
Move the repo_key to a Param, in order to keep support for older PE version and still can use the long key with the new apt module


electrical:
@cotocisternas thanks for the contribution. I'll let Jenkins do its magic and merge it when its green.

electrical:
jenkins, test it

electrical:
Would you be able to add a test for overriding the key id and source?

Thanks!

PR #507, opened 2015-11-30
Plugin Type & Provider

electrical:
Adding a custom Type and Provider for plugin installations


electrical:
jenkins, test it

electrical:
jenkins, test it

electrical:
jenkins, test it

electrical:
jenkins, test it

manifests/config.pp
```diff
@@ -65,10 +65,10 @@
       mode    => '0755',
     }
 
-    file { $elasticsearch::plugindir:
-      ensure  => 'directory',
-      recurse => true,
-    }
+    #file { $elasticsearch::plugindir:
```
rnelson0: This was added down below, the commented version should probably be removed.

lib/puppet/provider/elasticsearch_plugin/plugin.rb
```diff
@@ -0,0 +1,157 @@
...
+      @es_version = version.scan(/\d+\.\d+\.\d+(?:\-\S+)?/).first
+      debug "Found ES version #{@es_version}"
+    end
+  end
+
+  def is1x?
+    Puppet::Util::Package.versioncmp(@es_version, '2.0.0') < 0
+  end
+
+  def is2x?
```
rnelson0: For future proofing the API, this could check to make sure it's < 3.0.0 or possibly be renamed something like `isnot1x?`

manifests/config.pp
electrical: Good catch :-)

lib/puppet/provider/elasticsearch_plugin/plugin.rb
electrical: It could happen that 3.x might have other weird things requiring a `is3x?`
Not sure how to solve that in a nice way.
Any idea's?

rnelson0: `(Puppet::Util::Package.versioncmp(@es_version, '2.0.0') >= 0) and
(Puppet::Util::Package.versioncmp(@es_version, '3.0.0') < 0)` would
probably work to limit `is2x?`, then do similar for `is3x?` when it's out
and use that were possible. The question that I can't answer is would it be
better to then have to update other functions to say `if (x.is2x? or
x.is3x?)` or add additional if/else branches; or to rename the function and
change those instances now to `if x.isnot1x?`. Those may still need updated
after 3x is released anyway.

Or, since tons of code may need touched then anyway, fix `is2x?` to limit
`2.0.0<=@es_version<3.0.0` and worry about it later :)


Rob Nelson
rnelson0@gmail.com

lib/puppet/provider/elasticsearch_plugin/plugin.rb
```diff
@@ -0,0 +1,157 @@
...
+
+Puppet::Type.type(:elasticsearch_plugin).provide(:plugin) do
+  desc "A provider for the resource type `elasticsearch_plugin`,
+        which handles plugin installation"
+
+  commands :plugin => '/usr/share/elasticsearch/bin/plugin'
+  commands :es => '/usr/share/elasticsearch/bin/elasticsearch'
+
+  def exists?
+    es_version
```
igalic: this function should be called elsewhere, it doesn't make sense to call it in every sub-function that depends on it

lib/puppet/provider/elasticsearch_plugin/plugin.rb
```diff
@@ -0,0 +1,157 @@
...
+    if !File.exists?(pluginfile)
+      debug "Plugin file #{pluginfile} does not exist"
+      return false
+    elsif File.exists?(pluginfile) && readpluginfile != pluginfile_content
+      debug "Got #{readpluginfile} Expected #{pluginfile_content}"
+      self.destroy
```
igalic: why are we calling destroy here?

lib/puppet/provider/elasticsearch_plugin/plugin.rb
```diff
@@ -0,0 +1,157 @@
...
+      false
+    else
+      debug "Plugin exists"
+      return true
+    end
+  end
```
igalic: i'd restructure this whole function to
```ruby
def exists?
  return false unless File.exists?(pluginfile)
  if File.exists?(pluginfile) && readpluginfile != pluginfile_content
    warn "Destroying malformed plugin #{pluginfile}"
    debug "Got #{readpluginfile} Expected #{pluginfile_ccontent}"
    self.destroy
    return false
  end
  true
end

File: lib/puppet/provider/elasticsearch_plugin/plugin.rb [hunk 6]
```diff
@@ -0,0 +1,157 @@
...
+    f = File.open(pluginfile)
+    f.readline
+  end
+
+  def install1x
+    if !@resource[:url].nil?
+      commands = [ plugin_name(@resource[:name]), '--url', @resource[:url] ]
+    elsif !@resource[:source].nil?
+      commands = [ plugin_name(@resource[:name]), '--url', "file://#{@resource[:source]}" ]
+    else
```
igalic: this whole section reads really confusing.

from what i gather, `url` and `source` will be treated differently, when neither from a user perspective, nor a program perspective they are.

lib/puppet/provider/elasticsearch_plugin/plugin.rb
```diff
@@ -0,0 +1,157 @@
...
+
+  def install2x
+    if !@resource[:url].nil?
+      commands = [ @resource[:url] ]
+    elsif !@resource[:source].nil?
+      commands = [ "file://#{@resource[:source]}" ]
+    else
+      commands = [ @resource[:name] ]
+    end
+    commands
```
igalic: imo: there should be only one source, and that source can alternatively be a URL, or a filesystem path.
if it is the latter, but doesn't start with `file://`, we make it start with `file://`

that way, these two functions are simplified reduced to only two cases each.

lib/puppet/provider/elasticsearch_plugin/plugin.rb
```diff
@@ -0,0 +1,157 @@
...
+  end
+
+  def is2x?
+    (Puppet::Util::Package.versioncmp(@es_version, '2.0.0') >= 0) && (Puppet::Util::Package.versioncmp(@es_version, '3.0.0') < 0)
+  end
+
+  def plugin_version(plugin_name)
+    items = plugin_name.split("/")
+    if is1x?
+      if items.count == 3 # 'mobz/elasticsearch-head/1.2.3'
```
igalic: this whole bit can be done easier with destructuring:
```ruby
vendor, plugin, version =  'mobz/elasticsearch-head'.split('/')
```

lib/puppet/provider/elasticsearch_plugin/plugin.rb
```diff
@@ -0,0 +1,157 @@
...
+        return items[2].scan(/\d+\.\d+\.\d+(?:\-\S+)?/).first
```
igalic: suddenly we're dealing with [names](https://en.wikipedia.org/wiki/Connascence_%28computer_programming%29), not positions (see Jim Weireich's [Grand Unified Theory of Software](https://vimeo.com/10837903) :)

lib/puppet/provider/elasticsearch_plugin/plugin.rb
```diff
@@ -0,0 +1,157 @@
...
+      end
+    elsif is2x?
+      if items.count == 1
+        return @es_version
```
igalic: what? why? huh??

lib/puppet/provider/elasticsearch_plugin/plugin.rb
```diff
@@ -0,0 +1,157 @@
...
+      elsif items.count == 3
+        return items[2].scan(/\d+\.\d+\.\d+(?:\-\S+)?/).first
+      end
+    end
+    return false
+  end
+
+  def plugin_name(plugin_name)
+
+    items = plugin_name.split("/")
```
igalic: same as above plz

lib/puppet/provider/elasticsearch_plugin/plugin.rb
```diff
@@ -0,0 +1,157 @@
...
+
+    if items.count == 1
+      endname = items[0]
+    elsif items.count > 1
+      plugin = items[1]
+      if plugin.include?('-') # example elasticsearch-head
+        if plugin.start_with?('elasticsearch-')
```
igalic: why not `split('-')`?

lib/puppet/type/elasticsearch_plugin.rb
```diff
@@ -0,0 +1,31 @@
...
+    desc 'An arbitrary name used as the identity of the resource.'
+  end
+
+  newparam(:url) do
+    desc 'Url of the package'
+  end
+
+  newparam(:source) do
+    desc 'Source of the package. puppet:// or file:// resource'
+  end
```
igalic: as mentioned above, `:url` and `:source` should probably be unified

lib/puppet/type/elasticsearch_plugin.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  newparam(:proxy_args) do
+    desc 'Proxy Host'
```
igalic: this is maybe something that can be done through an environment setting…

lib/puppet/provider/elasticsearch_plugin/plugin.rb
```diff
@@ -0,0 +1,157 @@
+$LOAD_PATH.unshift(File.join(File.dirname(__FILE__),"..","..",".."))
+
+Puppet::Type.type(:elasticsearch_plugin).provide(:plugin) do
+  desc "A provider for the resource type `elasticsearch_plugin`,
+        which handles plugin installation"
+
+  commands :plugin => '/usr/share/elasticsearch/bin/plugin'
+  commands :es => '/usr/share/elasticsearch/bin/elasticsearch'
```
igalic: this means the commands will execute as root, is that okay so?

also: are there any scenarios in which these commands can be in a different location?

lib/puppet/provider/elasticsearch_plugin/plugin.rb
```diff
@@ -0,0 +1,157 @@
...
+    return @es_version if @es_version
+    begin
+      version = es('-v') # ES 1.x
+    rescue
+      version = es('--version') # ES 2.x
+    rescue
+      raise "Unknown ES version. Got #{version.inspect}"
+    ensure
+      @es_version = version.scan(/\d+\.\d+\.\d+(?:\-\S+)?/).first
+      debug "Found ES version #{@es_version}"
```
igalic: i think we can get rid of this debug statement

lib/puppet/type/elasticsearch_plugin.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  newparam(:plugin_dir) do
+    desc 'Plugin directory'
+    defaultto '/usr/share/elasticsearch/plugins'
+  end
+
```
igalic: we could make es_version a read-only value, and pull execute the provider call from here…

lib/puppet/provider/elasticsearch_plugin/plugin.rb
electrical: This is for the situation when we do an upgrade.
For example if the resource name has changed with the version number `plugin/1.0.0` to `plugin/1.0.1` it uninstalls the existing plugin and then installs the new one.
ES plugin thing doesn't do upgrades :-(

lib/puppet/type/elasticsearch_plugin.rb
electrical: Possibly, but the current define already supports it so wanted to keep the same support in there.

lib/puppet/provider/elasticsearch_plugin/plugin.rb
electrical: Very small chance they are in a different location, they are meant to be executed as root yeah.

lib/puppet/provider/elasticsearch_plugin/plugin.rb
electrical: Debug statements only show up when running with --debug mode, can be useful when things go wrong.

lib/puppet/provider/elasticsearch_plugin/plugin.rb
electrical: Tried to place it somewhere else but couldn't find the right spot.

lib/puppet/provider/elasticsearch_plugin/plugin.rb
electrical: So far i always treated source and url differently.
Source is used with a puppet:// resource and places the file in the /tmp folder.

lib/puppet/provider/elasticsearch_plugin/plugin.rb
electrical: Because i want to make sure it starts with 'elasticsearch' and not with anything else.

lib/puppet/type/elasticsearch_plugin.rb
electrical: Could do, will have to add some checks to it.

lib/puppet/type/elasticsearch_plugin.rb
electrical: How do you mean?

lib/puppet/provider/elasticsearch_plugin/plugin.rb
electrical: I wanted to have the debug statements in there at all parts so its easier to trace down issues when running with --debug flag.

lib/puppet/provider/elasticsearch_plugin/plugin.rb
electrical: Makes sense yeah

lib/puppet/provider/elasticsearch_plugin/plugin.rb
electrical: For 2.x plugins that only have a single name like `marvel` or `shield` which are classified as 'official plugins' the version of the plugin is the same as the ES version.
Because of that i use the es_version variable.

lib/puppet/provider/elasticsearch_plugin/plugin.rb
igalic: that seems like a bad place for them… /tmp is rather volatile ;)

lib/puppet/provider/elasticsearch_plugin/plugin.rb
igalic: we should still check if this is supported by all ruby versions you're targeting

lib/puppet/provider/elasticsearch_plugin/plugin.rb
igalic: ok. that code was very confusing to read… so i didn't know what it meant

lib/puppet/provider/elasticsearch_plugin/plugin.rb
```diff
@@ -0,0 +1,157 @@
...
+
+    if items.count == 1
+      endname = items[0]
+    elsif items.count > 1
+      plugin = items[1]
+      if plugin.include?('-') # example elasticsearch-head
+        if plugin.start_with?('elasticsearch-')
+          endname = plugin.gsub('elasticsearch-', '')
+        elsif plugin.start_with?('es-')
+          endname = plugin.gsub('es-', '')
```
igalic: or 'es-'

this is the rare case where i'd recommend using regex.

lib/puppet/type/elasticsearch_plugin.rb
igalic: read-only values in a type: https://github.com/puppetlabs/puppetlabs-apt/blob/master/lib/puppet/type/apt_key.rb#L72-L134

running a provider function from the type: https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/lib/puppet/type/postgresql_psql.rb#L23-L26

lib/puppet/provider/elasticsearch_plugin/plugin.rb
electrical: hmm yeah. good point.

lib/puppet/provider/elasticsearch_plugin/plugin.rb
electrical: 1.8.7 and up. Will do some checks.

lib/puppet/provider/elasticsearch_plugin/plugin.rb
electrical: it is yeah, i should add some comments :-)

lib/puppet/provider/elasticsearch_plugin/plugin.rb
electrical: Makes sense yeah, might be cleaner.

lib/puppet/type/elasticsearch_plugin.rb
electrical: Ahh okay. will try that.

electrical: No joy with this stuff so far. Might refactor it at some point when found how to solve it :-)

lib/puppet/provider/elasticsearch_plugin/plugin.rb
```diff
@@ -0,0 +1,130 @@
...
+
+  def exists?
+    es_version
+    if !File.exists?(pluginfile)
+      debug "Plugin file #{pluginfile} does not exist"
+      return false
+    elsif File.exists?(pluginfile) && readpluginfile != pluginfile_content
+      debug "Got #{readpluginfile} Expected #{pluginfile_content}. Removing for reinstall"
+      self.destroy
+      false
```
igalic: it's a bit confusing to have the difference between false / return false / return true…

lib/puppet/provider/elasticsearch_plugin/plugin.rb
igalic: the --debug flag should give you all of this for free

electrical: Only if i add the debug statements yeah, otherwise the logging can be confusing. at least when i started debugging things. that's why i added them.

PR #508, opened 2015-12-02
Export es_include

rhoml:
Export ES_INCLUDE variable ensuring this is pick up from the right default file otherwise it will only use /usr/share/elasticsearch/elasticsearch.in.sh

### DIFF
--- a/templates/etc/init.d/elasticsearch.Debian.erb
+++ b/templates/etc/init.d/elasticsearch.Debian.erb
@@ -111,6 +111,7 @@ export ES_HEAP_NEWSIZE
 export ES_DIRECT_SIZE
 export ES_JAVA_OPTS
 export ES_CLASSPATH
+export ES_INCLUDE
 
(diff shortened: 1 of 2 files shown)


electrical:
Could you add this to the RedHat init script as well so it works there as well?

Thanks!

rhoml:
@electrical done :)

electrical:
@rhoml thanks! Will let jenkins do its thing but i assume nothing will break :-)

electrical:
jenkins, test it

rhoml:
Brilliant! thank you

rhoml:
@electrical is this something I should fix?

electrical:
I'll have a look what's wrong when I get to the office. Seems something is wrong with one of the jenkins slaves.

electrical:
jenkins, test it

rhoml:
<3 thanks @electrical

PR #530, opened 2015-12-17
add Gentoo support

derdanne:
add Gentoo support

### DIFF
--- a/manifests/service/openrc.pp
+++ b/manifests/service/openrc.pp
@@ -1,0 +1,195 @@
+# == Define: elasticsearch::service::init
+#
+# This class exists to coordinate all service management related actions,
+# functionality and logical units in a central place.
+#
+# <b>Note:</b> "service" is the Puppet term and type for background processes
+# in general and is used in a platform-independent way. E.g. "service" means
+# "daemon" in relation to Unix-like systems.
+#
+#
+# === Parameters
+#
+# [*ensure*]
+#   String. Controls if the managed resources shall be <tt>present</tt> or
+#   <tt>absent</tt>. If set to <tt>absent</tt>:
+#   * The managed software packages are being uninstalled.
+#   * Any traces of the packages will be purged as good as possible. This may
+#     include existing configuration files. The exact behavior is provider
+#     dependent. Q.v.:
+#     * Puppet type reference: {package, "purgeable"}[http://j.mp/xbxmNP]
+#     * {Puppet's package provider source code}[http://j.mp/wtVCaL]
+#   * System modifications (if any) will be reverted as good as possible
+#     (e.g. removal of created users, services, changed log settings, ...).
+#   * This is thus destructive and should be used with care.
+#   Defaults to <tt>present</tt>.
+#
+# [*status*]
+#   String to define the status of the service. Possible values:
+#   * <tt>enabled</tt>: Service is running and will be started at boot time.
+#   * <tt>disabled</tt>: Service is stopped and will not be started at boot
+#     time.
+#   * <tt>running</tt>: Service is running but will not be started at boot time.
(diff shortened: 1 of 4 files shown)


manifests/package.pp
```diff
@@ -24,154 +24,163 @@
...
-      creates => $elasticsearch::package_dir,
-    }
-
-    file { $elasticsearch::package_dir:
-      ensure  => 'directory',
-      purge   => $elasticsearch::purge_package_dir,
-      force   => $elasticsearch::purge_package_dir,
-      backup  => false,
-      require => Exec['create_package_dir_elasticsearch'],
+  case $::osfamily {
```
electrical: Any specific reason for this? Why wouldn't the current code be able to handle it?

derdanne:
Gentoo has elasticsearch as native package provided by portage - i would prefer to manage the installation the native way.

electrical:
@derdanne by default it uses the native way for any distro, all the extra code is just optional when you want to use the package_url option, so you don't have to make it a case statement.

derdanne:
Oh, i see - reverted the changes of package.pp

manifests/service/openrc.pp
```diff
@@ -0,0 +1,195 @@
...
+      if(has_key($init_defaults, 'ES_USER')) {
+        if($init_defaults['ES_USER'] != $elasticsearch::elasticsearch_user) {
+          fail('Found ES_USER setting for init_defaults but is not same as elasticsearch_user setting. Please use elasticsearch_user setting.')
+        }
+      }
+
+      $init_defaults_pre_hash = { 'ES_USER' => $elasticsearch::elasticsearch_user, 'ES_GROUP' => $elasticsearch::elasticsearch_group, 'MAX_OPEN_FILES' => '65535' }
+      $new_init_defaults = merge($init_defaults_pre_hash, $init_defaults)
+
+      augeas { "defaults_${name}":
```
electrical: I think this block can be removed since the file isn't used anywhere? Or am i missing it?

derdanne: The defaults file is used:

$defaults_location  = '/etc/conf.d' in params.pp 
and "${elasticsearch::params::defaults_location}/elasticsearch.${name}" matches with init file /etc/init.d/elasticsearch.${name}

electrical: Ahh okay. i didn't see it in the openrc script. I guess its being read automatically then?

derdanne: Yes, the defaults file is automatically parsed, when it has the same name as the init script.

electrical: Ahh okay. much smarter then the init scripts :-)

electrical:
Thanks for all the work so far. I'll let Jenkins do its magic and merge it when its green.

electrical:
jenkins, test it

electrical:
jenkins, test it

electrical:
@derdanne would you be able to squash your commits into a single commit? Then i will merge it.
Thanks!

derdanne:
Can you give me a hint how i can revert the merges from master? 

I just pushed the github button "merge changes from master", and i don't get this working :/

electrical:
Ahh okay. you can do a `git rebase -i`  which gives you the option to pick and squash commits.
If you set the first one to pick and the other 2 to squash you can modify it all into a single commit.

derdanne:
OK, that does not seem to work - then i get all commits but not the merges - i think i have to do a new pull request with my changes only :(

electrical:
No worries :) if you could open a new pr then would be great.

Thanks.

derdanne:
Ok, i've created a new PR https://github.com/elastic/puppet-elasticsearch/pull/559

PR #539, opened 2016-01-06
Added ES_GC_LOG_FILE export in init files

### LINKED ISSUE #538: Support GC Logging option
Elasticsearch contains some logic to properly enable GC logging if ES_GC_LOG_FILE defined.

A "$gc_log_file" option in the module would be nice, that would then add `export ES_GC_LOG_FILE` if defined.

ntent-ashton:
Per the official distribution init.d files which contain this line:
https://github.com/elastic/elasticsearch/blob/8ffb63a177679141da65981c41340a929ed4ae2d/distribution/rpm/src/main/packaging/init.d/elasticsearch#L66
https://github.com/elastic/elasticsearch/blob/8ffb63a177679141da65981c41340a929ed4ae2d/distribution/deb/src/main/packaging/init.d/elasticsearch#L108

If this is in by default, it doesn't alter or harm anything.  It also allows me to add ES_GC_LOG_FILE to my init_defaults hash, and have it work without having to add a conditional to the template file.

Closes #538

### DIFF
--- a/templates/etc/init.d/elasticsearch.Debian.erb
+++ b/templates/etc/init.d/elasticsearch.Debian.erb
@@ -112,6 +112,7 @@ export ES_DIRECT_SIZE
 export ES_JAVA_OPTS
 export ES_CLASSPATH
 export ES_INCLUDE
+export ES_GC_LOG_FILE
 
 # Check DAEMON exists
 test -x $DAEMON || exit 0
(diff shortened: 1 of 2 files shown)


electrical:
Looks good to me yeah. Must admit i haven't updated the init / systemd files for a while.. really should do that some day :-)

ntent-ashton:
Thank you!

PR #545, opened 2016-01-08
Solve issues with ES being run

### LINKED ISSUE #528: puppet-elasticsearch freezes with ES 2.1 on CentOS 6
To determine the version of the installed instance of ES '/usr/share/elasticsearch/bin/elasticsearch -v' is started. On ES 1.7 it returns the version, on ES 2.x it just runs ES and it does not return to puppet.

As a result the puppet run freezes ...

When I remove the line 'version = es('-v') # ES 1.x' (and the 'rescue' line just below it) from the file lib/puppet/provider/elasticsearch_plugin/plugin.rb, puppet runs perfectly on my updated systems. Now it will freeze on systems not upgraded to 2.x though.

electrical:
In some weird cases it could happen that with the first command ES 2.x
would actually startup instead of fail.
This uses the version flag for ES 2.x but also works on 1.x which
ignores everything after -v

Closes #528

### DIFF
--- a/lib/puppet/provider/elasticsearch_plugin/plugin.rb
+++ b/lib/puppet/provider/elasticsearch_plugin/plugin.rb
@@ -90,9 +90,7 @@ Puppet::Type.type(:elasticsearch_plugin).provide(:plugin) do
   def es_version
     return @es_version if @es_version
     begin
-      version = es('-v') # ES 1.x
-    rescue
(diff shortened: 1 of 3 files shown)


electrical:
jenkins, test it

electrical:
jenkins, test it

electrical:
jenkins, test it

PR #554, opened 2016-01-14
Still create init script when status is unmanaged

### LINKED ISSUE #482: new issue - never had before -- centos 6.7 - init.d service not created (using 0.9.8)
I have a new problem going on I've never experienced before. Stood up some new VMs to create a new cluster using this module. Everything is getting created accept the init.d instance service. 

I'm using 0.9.8 module
PE 2015.2
ES 1.4.2
CentOS 6.7 64bit

no /etc/init.d/elasticsearch-[instance name] is getting created.. everything goes fine with puppet run, no visible errors. I've had mirrored setup without issues. 

So to try and bypass puppet altogether I tried to copy the service script from another box that works, edited it to match new instance name, and installed it as a new service under chkconfig. And when I tried to run it.. I get invalid runuser option --pidfile. Is this a sign that I am missing some update? Yum says I have none.

Here's  a copy of most recent run... you can see where it deletes the default init.d -- which is expected, but never creates the new init.d script with instance name:

http://pastebin.com/L3J1vqAi

### LINKED ISSUE #499: when using 'unmanaged' for service, the init script is not installed
Is there a particular reason why the init script isn't installed when using status=unmanaged (I'm using pacemaker to manage the elasticsearch service)?

### LINKED ISSUE #541: create init script but don't manage service when set to unmanaged
reported in https://github.com/elastic/puppet-elasticsearch/issues/499 and https://github.com/elastic/puppet-elasticsearch/issues/482
When the module or instance is set to unmanaged it should still create the init scripting but not enable or start it.

electrical:
When we set the service status to unmanaged we still need to create the
init script so they can manage the service via other tooling.

Closes #482
Closes #499
Closes #541

### DIFF
--- a/manifests/service/init.pp
+++ b/manifests/service/init.pp
@@ -113,7 +113,7 @@ define elasticsearch::service::init(
   }
 
 
-  if ( $status != 'unmanaged' and $ensure == 'present' ) {
+  if ( $ensure == 'present' ) {
 
     # defaults file content. Either from a hash or file
     if ($init_defaults_file != undef) {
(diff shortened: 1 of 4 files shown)


electrical:
jenkins, test it

electrical:
jenkins, test it

Issue #553, opened 2016-01-14
should support optionally not managing $datadir


alexharv074:
Some people will be using LVM, mounted SAN storage etc for the ElasticSearch data directory, whereas this module assumes that it can make the data directory.  There needs to be a feature to optionally disable this.

bflad:
For what its worth, we generally set the elasticsearch data directory one subdirectory below anything we LVM/mount.

alexharv074:
While there's nothing wrong with that, I still think the module should support use of a directory that was created outside of the module.

electrical:
Hi guys,

I've added the discussion label to get more input from others before making a decision on this.

alexharv074:
@electrical well to get things moving I already submitted a pull request.

Please note that I am upgrading from both an earlier version of ElasticSearch and an earlier version of this module.  Seems like the earlier version of this module evidently didn't try to manage the $datadir?  Unless my colleague hacked it. :)

To me it's a no-brainer, of course the module should allow people to manage the LVM externally to the module, and some people simply won't be able to use a subdirectory.  I can't, if the feature can't be added I'll simply have to fork it.

PR #556, opened 2016-01-14
Add feature $manage_datadir

alexharv074:
See discussion in #553

### DIFF
--- a/manifests/instance.pp
+++ b/manifests/instance.pp
@@ -78,7 +78,7 @@ define elasticsearch::instance(
   $status             = $elasticsearch::status,
   $config             = undef,
   $configdir          = undef,
-  $manage_datadir     = true,
+  $manage_datadir     = $elasticsearch::manage_datadir,
   $datadir            = undef,
   $logging_file       = undef,
   $logging_config     = undef,
--- a/manifests/instance.pp
+++ b/manifests/instance.pp
@@ -42,6 +42,11 @@
 #   Path to directory containing the elasticsearch configuration.
 #   Use this setting if your packages deviate from the norm (/etc/elasticsearch)
 #
+# [*manage_datadir*]
+#   Whether or not this module should create the datadir/s.  This allows you to
+#   create the directories for instance using LVM outside of this module.
+#   Defaults to <tt>true</tt>.
+#
 # [*datadir*]
 #   Allows you to set the data directory of Elasticsearch
 #
@@ -73,6 +78,7 @@ define elasticsearch::instance(
   $status             = $elasticsearch::status,
   $config             = undef,
   $configdir          = undef,
+  $manage_datadir     = true,
   $datadir            = undef,
   $logging_file       = undef,
   $logging_config     = undef,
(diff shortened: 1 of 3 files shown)


manifests/instance.pp
```diff
@@ -73,6 +78,7 @@
   $status             = $elasticsearch::status,
   $config             = undef,
   $configdir          = undef,
+  $manage_datadir     = true,
```
electrical: Might be easier to have this one default to $elasticsearch::manage_datadir ?
Then you also don't need the logic further down.

alexharv074: Are you sure?  If I have it default to $elasticsearch::manage_datadir, I will still have the problem that the user may override the default either in the elasticsearch class, or in the elasticsearch::instance type.  Of course, I'd rather just force to user to set it in the elasticsearch::instance type, however it appears to me from the design that the preference is to allow the user to set things in two places.

electrical: When using $elasticsearch::manage_datadir as a default value for $manage_datadir in the instance define it allows you to either set it to false for every instance you create, or you can override it per individual instance.
This way if the user has 2 or more instances, the value doesn't have to be duplicated for each instance.

alexharv074: Got it, I see what you're saying.  Let me fix that.

alexharv074: OK, updated.

manifests/instance.pp
```diff
@@ -112,6 +118,14 @@
     $instance_configdir = $configdir
   }
 
+  # manage_dir setting
```
electrical: You can get rid of this logic now with the modification you made for the default value :-)

alexharv074: Alright, unless I'm going slightly 1.30am crazy.

The user may choose to set $manage_dir to false using:

```
elasticsearch::instance { 'es-01':
  manage_datadir => false,
}
```

In that case, $elasticsearch::manage_datadir will be true and $manage_datadir will be false, and we'll fall through to the 'else'.

Still not seeing how I can remove this, as much as I wish I'd not had to put it in in the first place.

alexharv074: Ah, sorry.  It is the 1.30am crazy.  Ok, I'll fix it in the morning. :)

electrical: Haha that's okay :-)

electrical:
jenkins, test it

manifests/instance.pp
alexharv074: Right, fixed.

alexharv074:
Hi @electrical what is the status here.  Do you have any remaining concerns about this feature?

electrical:
could you rebase the PR? it seems you got a commit of mine in your PR.

alexharv074:
@electrical  Fyi I have refactored all the code here to use a config as recommended in the puppetlabs/lvm module.  After doing this, I have found that I no longer have the need for this feature.  Would you still like to merge this PR in anyway?

alexharv074:
Anyhow I have rebased.

alexharv074:
Actually, if combining with standard puppetlabs/lvm you would need:
```
$datadir = '/srv/es'
class { 'elasticsearch':
  datadir => $datadir,
}
Mount[$datadir] -> File[$datadir]
```

electrical:
Hi,

If it all works without this modification i would rather not merge this in.
Feels it could cause confusion for people and might lead them in using other parts in a wrong way.
Thank you though for the contribution!

PR #561, opened 2016-01-19
Manage elasticsearch templates with hiera binding

Vincent--:
Allow use of elasticsearch::templates inside hiera:

``` YAML

---
elasticsearch::templates :
    templatename :
        content :
            template: "*"
            settings:
                number_of_replicas: 0
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,6 @@ elasticsearch::instance { 'es-01': }
 This will set up its own data directory and set the node name to `$hostname-$instance_name`
 
 ####Advanced options
-
 Instance specific options can be given:
 
 ```puppet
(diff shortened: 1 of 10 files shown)


electrical:
Hi, Would you be able to rebase your commits so the 2 merge commits don't show up?
After that I'll trigger jenkins to do a test.

Vincent--:
Done :-)

electrical:
Looks like the initial acceptance tests failed. See http://puppet.ci.elastic.co/view/Elasticsearch/job/puppet-elasticsearch-pr-acceptance/14/ for details.

electrical:
My best guess is that it doesn't do sorting correctly when calling to_json causing the order to be different on each run.
Could you try that out?

Vincent--:
You're right, this issue seems to be related to json order...

On my setup, I use ruby 1.9, that explains why it works : 
https://github.com/flori/json/issues/66

electrical:
I was afraid of that. I had the same issue with the yaml parts. Will have to think up a solution based on the code I wrote for the yaml file.

Vincent--:
Could we add this requirement to your module ?
https://gist.github.com/aj-jester/e0078c38db9eb7c1ef45

Any thoughts about another way ?

electrical:
I don't think there is much of an other way then doing it like that.
I was thinking about making my yaml script into a function as well, perhaps we could build a simple module that has both of them so people can use it if needed?
If you are open to that would you mind if we put it in the https://github.com/voxpupuli namespace?

tylerjl:
Apologies for the delayed action here, but 55c65a7f exposed every defined type/resource as a top-level parameter to the `elasticsearch` class so this should be possible now for templates or any other type (plugins, scripts, et cetera).

PR #562, opened 2016-01-20
restart services after packages updates

azhurbilo:


electrical:
Hi,

In what situation do you need this notification? I haven't seen any issues with this so far.

azhurbilo:
When I upgrade Elasticsearch from 1.7 to 2.1 for example, services will not restarted and old version (1.7) continue working

electrical:
Hi,

When setting the version variable to specify the version works fine, when its not set it only ensures the package to 'present'

azhurbilo:
> When setting the version variable to specify the version works fine, when its not set it only ensures the package to 'present'

Right! 
But I talk about case when we provision the same server/VM. 

Step to reproduce:
1. bootstrap new VM
2. provision "puppet-elasticsearch" with `version => '1.7.0',`
3. provision "puppet-elasticsearch" with `version => '2.1.0',`

Actual result: version '1.7.0' continue working

Expected result: all elasticsearch services restarted, version '2.1.0' is working

electrical:
That's weird because that should work fine as expected.
I actually have a test that does an upgrade test and works fine.

electrical:
I'll try to do some more digging around it this week.
Could you gist the whole puppet part you use for the puppet module? Thanks!

electrical:
Uhg. found the problem and you are right.
Reason that my tests pass is that other resources send a notify to the service :-(

electrical:
are you able to sign the CLA so i can merge it? thanks!

electrical:
I tried your fix but its in a part that only works if the `package_url` option is used. I moved it a bit higher and now it always notifies the service if needed.
I'll open a new PR and link it here so you can check it out.

electrical:
Please have a look at https://github.com/elastic/puppet-elasticsearch/pull/564

azhurbilo:
@electrical  :+1:  #564 cool! You can close this PR.
Strange that I signed the CLA, but continue get message here "Commit author has not signed the CLA".

electrical:
yeah. the CLA check thing is a bit weird at times. Closing this PR.
Thank you very much for raising this issue in the first place and coming up with a solution!

PR #563, opened 2016-01-25
Remove plugin dir when the package changes

electrical:
For plugins we need to remove them all before we can install new ones.
This is a temporary solution until something better comes up.


electrical:
jenkins, test it

PR #581, opened 2016-02-06
Issue_580  Stop package_dl_timeout interfering with global resource defaults

alexharv074:
If you're happy with this contribution, I'll make the same change in the logstash and logstash forwarder modules and any others that have the same issue.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -89,7 +89,7 @@
 #
 # [*package_dl_timeout*]
 #   For http,https and ftp downloads you can set howlong the exec resource may take.
-#   Defaults to: 600 seconds
+#   Defaults to: undef (Puppet's default of 300 or a resource default will be used).
 #
 # [*proxy_url*]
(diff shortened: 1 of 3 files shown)


electrical:
Hi, the 600 seconds is intentionally because with testing the 300 seconds was to short.

alexharv074:
I'm not surprised because on slow networks I'm finding 600 seconds is also too short.  On a normal network, though, I would expect you'll download the ES package in less than 300 seconds (my home network is ~ 2.45 and it's not that fast).

Don't you think allowing people to disable the timeout globally for all Execs in site.pp is a cleaner solution though?  If you've had tests that are failing, you could easily do what I'm proposing in your tests too.

Another option would be to change the timeout to 0 instead of 600, disabling the timeout altogether.

It's not a big deal if you don't want to do this; I can just use your timeout parameters on all the Elastic.co modules.

alexharv074:
Probably best to close this.

Issue #480, opened 2015-11-04
debian repo key too short


richardrasu:
Hi,

We see the following message in our puppet logs: Add key: D88E42B4 from Apt::Source elasticsearch]) The id should be a full fingerprint (40 characters), see README.

In the case of elasticsearch, the full fingerprint should be: 46095ACC8548582C1A2699A9D27D666CD88E42B4

Would it be possible for you to update the manifest (repo.pp), or do you require us to fork & PR?

Cheers

Richard

electrical:
Hi,

This is a known issue but because I'm still supporting Puppet Enterprise 3.7.x which runs with an older pl-apt version i'm unable to update it to use the longer fingerprint.

richardrasu:
OK, thanks for the quick response. I understand your predicament :(
Cheers,

R.

Issue #445, opened 2015-09-04
apt_key warning


arkadicolson:
Warning: /Apt_key[Add key: D88E42B4 from Apt::Source elasticsearch]: The id should be a full fingerprint (40 characters), see README.

Change D88E42B4 into 46095ACC8548582C1A2699A9D27D666CD88E42B4 in the repo.pp

Thx!

electrical:
Hi,

This is a known issue indeed. This is caused by using the short key which is required for some of the older puppet Enterprise versions i'm supporting which are pinned using older apt module versions.
As soon as i can drop support for those i'll update it, but figured a warning is better then dropping support for things :-)

PR #585, opened 2016-02-16
Update init.pp

wazoo:
Puppet apt package produces a warning when the full GPG key ID is not provided.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -240,7 +240,7 @@ class elasticsearch(
   $java_package          = undef,
   $manage_repo           = false,
   $repo_version          = undef,
-  $repo_key_id           = 'D88E42B4',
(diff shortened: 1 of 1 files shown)


wazoo:
I did sign the CLA after I submitted this, forgot to do it beforehand, apologies.

electrical:
the short key is used because in some older pl-apt versions it doesn't support the long version of the key.

tylerjl:
Thanks for initiating the work on this @wazoo - there's another PR in #607 that is trying to work around the warning while handling the versions of PE that can't handle the short version (as mentioned in [other](https://github.com/elastic/puppet-elasticsearch/issues/445) [issues](https://github.com/elastic/puppet-elasticsearch/issues/480)) so I'll close in favor of that one, hopefully we can get rid of the warning without breaking compatibility with some older PE versions.

PR #599, opened 2016-03-12
Issue_598  fix versioncmp calls for Puppet 4

alexharv074:


alexharv074:
Sorry, I realised the bug is elsewhere, apparently these facts are already strings in Puppet 4.

PR #600, opened 2016-03-16
Changed repo_key_id in init.pp

MarieHo:
Changed default repo_key_id value by fingerprint value to avoid puppet warnings like:
'/Apt_key[Add key: D88E42B4 from Apt::Source elasticsearch]
The id should be a full fingerprint (40 characters), see README.'


tylerjl:
Duplicate of #585

PR #607, opened 2016-03-25
Fix #606: Use short fingerprint only for Puppet Enterprise.

### LINKED ISSUE #606: Use 40 char repository key fingerprint.
Hi,

i know that this has already been requested in #445 and #480, but what about using the is_pe fact from puppetlabs/stdlib to differenciate between Puppet and Puppet Enterprise? You could even use pe_version, pe_major_version, pe_minor_version and pe_patch_version to decide when to use the short key fingerprint.

Would that be an option for you?

Cheers,
Fabian

onibox:


tylerjl:
Thank you for the PR @onibox - just wanted to check and let you know I'm working to add some testing across different versions of puppet enterprise to ensure this works across them. According to some old comments it sounds like PE 3.7.x uses the old pl-apt version, so we may be able to add a conditional about the Puppet Enterprise version in use to use the longer key to better support forward compatibility.

tylerjl:
Sorry for the long wait @onibox - could you merge in the latest changes from `master` so the tests will re-run with the complete suite of PE versions in Travis? That should give us a pretty comprehensive set of results against different versions of PE, Puppet, etc.

onibox:
@tylerjl Sorry for the late reply too. Did merge master into branch.

tylerjl:
Interesting; the [rspec tests failed](https://travis-ci.org/elastic/puppet-elasticsearch/jobs/135189768), they're choking on the `$::is_pe` variable. Not sure what's causing it, feel free to push a change if you find a fix, I'll also look when I get some time as well.

onibox:
@tylerjl although i found the issue using $::is_pe i would rather like to use $::pe_version for the condition. Do you know which puppet enterprise version supports the long fingerprints?

tylerjl:
I don't - but I think I'm probably going to follow the [maintenance lifecycle the Puppet Labs has outlined](https://puppet.com/misc/puppet-enterprise-lifecycle) and pull the tests for 3.7 in July, so if all the tests pass for >= 3.8, then I think we're probably safe.

ctavan:
Any updates on this one?

tylerjl:
It looks like `pe_version` is undefined on recent versions of Puppet Enterprise, which is causing tests on those versions of PE to fail.

tylerjl:
Thanks for doing the work here @onibox - Puppet (Inc.) has marked really old versions of PE as unsupported now (see their [support matrix](https://puppet.com/misc/puppet-enterprise-lifecycle)) so this should be pretty safe to merge even without the checks you've included.

I've forked off of your branch and submitted a new PR that doesn't require the `$::pe_version` check that was causing test failures; I can carry the rest of the work through to the next release. 👍  (let me know if you want to be included in `CONTRIBUTORS`, I don't want to steal your thunder if you wanted credit.) 😄

onibox:
@tylerjl Many thanks! I'd be happy getting included in `CONTRIBUTORS` as i'm planning more PR in the future :-) 🏆

PR #608, opened 2016-03-31
New Parameter homedir for class elasticsearch. Analogous to datadir, …

marcIhm:
…etc. E.g. allows to have no directory /usr/share/elasticsearch to be created. Default still comes from param.

Patch needed, because I was unable to move the whole installation from /usr/share/elasticsearch to /opt/xyz/elasticsearch; /usr/share/elasticsearch with some (mostly empty) subdirectories would always be created. Digging deeper I found, that most of the Manifests used the variable $elasticsearch::params::homedir, which can not be influenced by the user of this module and has a value of /usr/share/elasticsearch. My change simply replaces $elasticsearch::params::homedir with $elasticsearch::homedir throughout, the latter beeing a parameter to the class elasticsearch. Its default however comes from $elasticsearch::params::homedir, so that this change should be fully backward compatible.


tylerjl:
Hey @marcIhm, thanks for the PR! This looks very useful. A few notes:

- You'll need to [sign the CLA](http://www.elasticsearch.org/contributor-agreement/) before we can accept the PR
- The new parameter is [undocumeted](https://travis-ci.org/elastic/puppet-elasticsearch/jobs/119819244#L404) according to the Travis documentation tests
- There are dependent path variables that reference `/usr/share` (such as [this one](https://github.com/marcIhm/puppet-elasticsearch/blob/0942fecb14186cd791b0b8600b7856c52c9322c6/manifests/params.pp#L113) that should probably be based off the `$homedir` to remain consistent
- It would also be good to test an alternative homedir similar to the way [testing an alternative user is done](https://github.com/marcIhm/puppet-elasticsearch/blob/0942fecb14186cd791b0b8600b7856c52c9322c6/spec/classes/000_elasticsearch_init_spec.rb#L299) in the rspec tests

marcIhm:
Okay, very good !
- I did assign the CLA yesterday, but maybe the process takes more time
- Will look in the issue, you pointed at 
- Will add a test

tylerjl:
Thanks @marcIhm! Just a quick follow-up, the automated CLA check did pass this time, so you're good to go there.

marcIhm:
Hi Tyler,

good to hear, that my CLA has arrived. However, I have problems with rspec :-/
Admittedly I am new to rspec and puppet-rspec. On the other hand I have written regression-tests before and I know ruby. So I was optimistic to be able to run your tests and add my own. However, I tried to run your existing tests on friday but without success; in fact I spent considerably more time on testing than on modifying your code. Therefore my question: Do you have some guidelines on how to setup a system to run your tests ? Maybe a site.pp that I could use or similar ...

And sorry, if this should be too newbie a question !

regards
Marc

tylerjl:
No problem Marc, part of the work I've been doing is to make this project easier for people to test, so if you're having difficulty, it's a bug. :smile: 

[CONTRIBUTING.md](https://github.com/elastic/puppet-elasticsearch/blob/master/CONTRIBUTING.md) has detailed instructions regarding how to test, but the short version is that as long as you have **make** and **ruby** available, you *should* be able to just `make test-intake` at the root of the repository to run through some tests (the `test-intake` target runs the tests that are a part of `test-rspec` as well as documentation tests.) There are other tests that run in `make test-acceptance` but those are heavier, so we can let Travis worry about spending time on those.

The results that you see from `make test-intake` should tell you about undocumented parameters and all other tests that don't include spinning up a VM to run more detailed acceptance tests.

marcIhm:
Thanx. I will try !

marcIhm:
Could run your test successfully (easy with a fresh install); now I will add one test.

karmi:
Hi @marcIhm, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/puppet-elasticsearch/pull/608.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

marcIhm:
Hi Tyler,
In case you are wondering about my pull-request: I am still not done. This is because I have discovered, that the literal string '/usr/local/elasticsearch' appears throughout most of the provider (ruby-)classes for your custom types (e.g. elasticsearch_shield_user). And of course, such literal values need to be replaced by references (some way or another) to the Parameter homedir of the Puppet-Class elasticsearch. This replacement is easy in the context of an instance of the provider class (there is the "resource"-function just for that), but it is not easy in the context of the provider class itself; as an example have a look at for lib/puppet/provider/elasticsearch_shield_user/parsed.rb, line 17: "confine :exists => users". Currently I am pondering about this (as my time budget permits) and I will talk to a co-worker beginning of July. So please be patient with my pull-request ! But of course, and thats another reason why I am writing this, it would be great to have your opinion on this too !
regards
Marc

tylerjl:
No worries @marcIhm, your effort is greatly appreciated, I'm thrilled to see you're taking the time to do it right!  ✅ 

I do think the general approach you're taking is the right one: providing a default parameter on the type that the provider references through `resource` is probably the right way to go. It's pretty verbose in the fact that you've got to define the `homedir` parameter everywhere, but at least it'll be easier for users in that we can centralize it in the `params` class if it ever needs to get changed.

I'm assuming, based upon what you wrote, that using `resource` in the `confine` statement doesn't work since it needs an instance of the provider. If that's the case I'll think on it as well.

Thanks again for working on this!

marcIhm:
Yup, thanx, then we have the same view on this; I'll keep you posted ...
regards
Marc

manifests/init.pp
```diff
@@ -383,6 +435,7 @@
   $api_ca_path             = undef,
   $validate_tls            = true,
   $system_key              = undef,
+>>>>>>> a93350b4b13ad5c8f795c67a3ca3046a88cdf25a
```
tylerjl: Looks like there's some leftovers from merge conflicts

templates/etc/init.d/elasticsearch.RedHat.erb
```diff
@@ -31,7 +31,7 @@ if [ -f /etc/rc.d/init.d/functions ]; then
     . /etc/rc.d/init.d/functions
 fi
 
-exec="/usr/share/elasticsearch/bin/elasticsearch"
+exec="<%= homedir %>/bin/elasticsearch"
```
tylerjl: These interpolations probably won't work unless a top-level `$homedir` exists in `manifests/service/init.pp` - it can just mirror `$elasticsearch::homedir`, but it needs to be brought into scope so the template can interploate it.

marcIhm:
Hi Tyler,
Tried to merge and noticed, that around june a parameter homedir has been introduced (which is something I have tried too). 
So I think its better to start from the latest version and reapply my changes manually. I will do a new fork and pull request soon (today ?). I just keep this one to still have your comments.
Thanx for your support. Will be back soon.

marcIhm:
Hi Tyler,

just wanted to apologize for dumping my pull request to your module so 
abruptly !
Unfortunately my project will be able to use your module as is, and so 
they decided not to let me work any longer on this :-(

Thanx for your great support and encouragement !

best reards
Marc

Am 05.10.2016 um 21:59 schrieb Tyler Langlois:

PR #621, opened 2016-05-02
Syncing with upstream

dforste:
As part of your pull request, please review the following tasks:
- [ ] Rebased/up-to-date with base branch
- [ ] Tests pass

Syncing Template with upstream. Need to pay special attention to the Enviornment section. Otherwise a lightly modified copy of https://github.com/elastic/elasticsearch/blob/v2.3.2/distribution/src/main/packaging/systemd/elasticsearch.service


tylerjl:
It looks like the tests failed on most systemd-based distros, I didn't dive in too deep to the errors but the one closest to the surface is what gets returned when trying to start the service:

```
Failed at step EXEC spawning /usr/share/elasticsearch/bin/elasticsearch-systemd-pre-exec: No such file or directory
```

This executable may be present on some versions of ES but it will need to be conditional since it isn't on all versions.

tylerjl:
This LGTM, lots of these changes are definite 👍  to the unit.

One note here is that by nuking `Type=forking` and removing `-d` from `ExecStart`, the unit will **no longer** work for versions of Elasticsearch that defaulted to double-forking instead of foregrounded processes by default. I believe this is limited to versions of Elasticsearch < 1.0, though I'd like to confirm that with an ES dev - @nik9000, maybe you know?

That means we can either 1) move forward and officially deprecate 0.9 versions of ES for this module, or 2) keep `Type=forking` and pass `-d` in `ExecStart` (later versions respect it I believe, and 0.9 will eat the unknown option and continue to fork.) If we do 2), @dforste may have to re-confirm this fixes #622.

nik9000:
> @nik9000, maybe you know?

Looks like this is the change that switched it:
```
commit c30945a3d84c7e6aa516f123a140d344f5cec42e
Author: Alexander Reelsen <alexander@reelsen.net>
Date:   Sun Dec 15 20:29:11 2013 +0100

    Start elasticsearch in the foreground by default
    
    Instead of using the '-f' parameter to start elasticsearch in the
    foreground, this is now the default modus.
    
    In order to start elasticsearch in the background, the '-d' parameter
    can be used.
    
    Closes #4392
```

Looks like the first version that was released with was 1.0.0-rc1.

I do wonder if we should be using the systemd conf file that comes with Elasticsearch if possible. We try to make it useful.

tylerjl:
Thanks for the info @nik9000. Leveraging the existing unit that comes bundled with Elasticsearch is a definite possibility; I think the main reason the module ships its own is to avoid having to alter the bundled unit for per-instance services, but there may be better ways around it. At least this PR brings it up-to-date with what's current upstream.

@dforste would you be willing to try the unit with Type=forking and the -d flag and see if it still fixes your problem? Then we can maintain compatability until we hit something really unavoidable. If it doesn't work, I'm on-board moving ahead anyway since this fixes something important at the expense of a really old version.

Xylakant:
systemd drop ins would be a way to leverage the existing unit file while still overwriting certain aspects of the config. Given the package-provided unit file in `/usr/lib/systemd/system/elasticsearch.service` it's possible to symlink it to `/etc/systemd/system/elasticsearch-servicename` and provide a drop-in `/etc/systemd/system/elasticsearch-servicename.d/drop-in.conf` that overwrites the relevant environment vars and configs. (see https://www.freedesktop.org/software/systemd/man/systemd.unit.html#id-1.11.3 for an example)

However, the problem might be that different versions of the ES package use different systemd unit files and thus may require different drop-in files: Environment vars defined via `EnvironmentFile=` in an env file can only be overwritten with a new EnvironmentFile while Environment vars defined directly in the unit file via `Environment=` can be overwritten both by a new `Environment=` statement and by Environment files. (see https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=)

tylerjl:
@Xylakant I really like that idea. Part of the reason this PR is an issue is the discrepancies between differences in unit files for different versions of elasticsearch. Symlinking the vendor unit into an instance unit keeps the instance consistent with what the package shipped, allows us to override certain parameters, and so on.

This wasn't possible until recently; #644 now masks the vendor unit so that it's hidden (consistent with the other service file behavior in this module to avoid confusion) but it should still exist so it can be symlinked to instance services.

The only problem I see here is backward compatability: if we symlink instance services from the vendor service, everyone on previous versions of the module won't have the vendor unit since it used to be `ensure => absent`. Not sure how to get around that one without some pretty annoying steps in the release notes (i.e. restore the file, reinstall the package, etc). Currently thinking on better ways to handle it.

tylerjl:
These updates (along with necessary checks to change `-Des.` to `-E` flags as necessary) have been merged in #785. Switching to systemd drop-ins is still a good idea, but the work can go on in another ticket.

PR #624, opened 2016-05-05
Shield Support: First Draft

tylerjl:
There's a lot here - start at the updated section in the README (look under "Advanced features"), that's a good place to start.

I'm particularly interested in user feedback regarding the implementation to see if there are important parts I'm missing or if there are different approaches people would prefer from a user perspective.

This **could** be merged right now (there are *lots* of tests and they all pass), though I'd like to add the following if I have time:

- [x] Add another provider for users such that people can control user passwords with hashes generated from `esusers`, alleviating the need to define passwords in plaintext (this seems to be the de facto way to encode passwords if you look at the default [users](https://docs.puppet.com/puppet/latest/reference/type.html#user-attribute-password) resource, for example.)

Important notes:

- This PR does **not** implement the ["native" Shield realm](https://www.elastic.co/guide/en/shield/current/native-realm.html). I punted on this for now since no GA releases are missing `esusers`, though this can be added with a provider pretty easily. (do note that the `native` provider means that you can't use hashed passwords in manifests, though.)
- Defining plaintext passwords requires a refresh on the type when changing Shield user passwords.
- Shield config files are centralized in `/usr/share/elasticsearch/shield` and synced via a `file` resource into instance directories in `/etc/elasticsearch/$instance/shield`. This is done because 1) symlinks don't work with the ES security manager, 2) hard links are problematic, and 3) controlling each instance's shield files would require a **lot** of overhead and complexity.

Feedback appreciated!  🎲


fgimian:
Hop @tylerjl, this looks really great!  Just one question, how will Shield know which instances(s) to apply the users / roles to?

tylerjl:
@fgimian right now when you install an Elasticsearch plugin, the plugin is shared across all instances of Elasticsearch running on the machine. For that reason, users and roles are shared as well (installing the Shield plugin in puppet but then not applying the users/roles to some instances would just render the instance completely closed off.) Hope that makes sense.

fgimian:
Thanks @tylerjl, yeah it does mate.  But I can't seem to see where you can specify the instances where file-based users and roles should be applied.

e.g. these should be duplicated in

```
/etc/elasticsearch/instance1/shield
/etc/elasticsearch/instance2/shield
/etc/elasticsearch/instance3/shield
...
```

Perhaps I'm missing something?

tylerjl:
@fgimian the module handles keeping the shield directories in sync, check out the diff in the `instances` defined type.

fgimian:
@tylerjl Nice mate!  Well I may just try this branch out.  I just finished my own implementation of this but it's very primitive by comparison.  If I have some time today, I may try your pull request on our build to get a real-world experience with it.

Huge huge thanks for your efforts on this! 😄

README.md
```diff
@@ -354,7 +356,116 @@ class { 'elasticsearch':
 
 Note: `init_defaults` hash can be passed to the main class and to the instance.
 
-##Advanced features
+## Advanced features
```
jarpy: Excellent docs! :+1:

spec/unit/provider/elasticsearch_shield_role/parsed_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      context 'with no roles' do
+        it 'should return no resources' do
+          expect(described_class.parse("\n")).to eq([])
+        end
+      end
+
+      context 'with one role' do
+        it 'should return one resource' do
+          expect(described_class.parse(%q{
+admin:
```
jarpy: Could I persuade to indent these heredocs? If you care about having no leading whitespace, you could use a method on the heredoc, like this one using the (trivial) `unindent` Gem.

```
[19] pry(main)> heredoc = <<-EOF.unindent
[19] pry(main)*   stuff:
[19] pry(main)*     - more
[19] pry(main)*     - stuff
[19] pry(main)*   EOF
=> "stuff:\n  - more\n  - stuff\n"

[20] pry(main)> puts heredoc
stuff:
  - more
  - stuff
=> nil
```

spec/acceptance/018_shield_spec.rb
```diff
@@ -0,0 +1,306 @@
...
+EOF
+      end
+
+      it 'should apply cleanly' do
+        apply_manifest single_manifest, :catch_failures => true
+      end
+
+      it 'should be idempotent' do
+        expect(apply_manifest(
+          single_manifest, :catch_failures => true
```
jarpy: You can also do `:catch_changes => true`, so you don't have to check the exit code.

spec/acceptance/018_shield_spec.rb
```diff
@@ -0,0 +1,306 @@
...
+    end
+
+    describe "secured REST endpoint" do
+      it 'authorizes changed passwords' do
+        curl_with_retries(
+          'elastic user cluster health request with new password',
+          default,
+          "-s -I -XGET -u #{@user}pwchange:#{@user_password[0..5]} "\
+          "http://localhost:9200/_cluster/health " \
+            "| grep '200 OK'", 0)
```
jarpy: It looks like you could use a helper method around `curl_with_retries` (or maybe just curl). It's hard to see the intent of this test because of all the boilerplate you have to feed the function call.

Also, passing `| grep` as a part the `url` argument is a clever hack, but it's kinda nasty.

jarpy:
Awesome to see high quality acceptance tests. :rocket:

spec/acceptance/018_shield_spec.rb
tylerjl: Thanks for the other excellent suggestions @jarpy, was not aware of some of those. 👍 

Regarding `grep` piping, I completely agree and actually attempted to circumvent this before falling back to plain shell piping; see 673aa7f for a failed experiment using [infrataster](https://github.com/ryotarai/infrataster) for a more structured approach using Faraday. I ended up being unable to make it work, however.

Do you have another suggestion to make this less janky? I'm no rspec expert so I may be missing an obvious solution.

jarpy: First off, don't let my comments stop you merging. Working tests + working code == :star:.

I also think using curl is fine. The tests are Situated Software, and we can expect them to be running in a sane Linux environment with curl available. I just think it's the API of `curl_with_retries` that is slowing you down.

I think having our own helper method could lead to much cleaner tests. Something along the line of this perhaps?

``` ruby
describe 'example.org' do
  def status_of(url)
    `curl -s -w '%{http_code}' -o /dev/null #{url}`.to_i
  end

  it 'is http ok' do
    expect(status_of('http://example.org')).to eq(200)
  end
end
```

I'm happy to have a hack on this later if you like :)

fgimian:
@tylerjl Forgive my lateness, I have just installed this module via r10k now and will be testing it out thoroughly tomorrow as I need to manage a few more file-based users via Shield.

I'll be in touch soon with an update
Fotis

fgimian:
OK, I have a few questions:

1. How would you represent the following role:

    ```yaml
      logstash:
        cluster:
          - manage_index_templates
        indices:
          - names:
              - 'logstash-*'
              - 'myapp-*'
            privileges:
              - write
              - delete
              - create_index
    ```

    Would I need to repeat the privileges for 2 keys?

2. Can I specify a role mapping for LDAP users?

    e.g.

    ```
    power_user:
      - "cn=group123,ou=group,ou=devteam,dc=aaa,dc=example,dc=com"
    ```

Apart from those points, the module appears to be working correctly for me so far :smile: 
Fotis

fgimian:
Just found the answer to the 2nd question in the code, using the mappings variable.  Perhaps add an example like that to the README as well? :smile:

fgimian:
And answer to the 1st question is to explicitly specify names and privileges:

e.g.

```
  elasticsearch::shield::role { 'logstash':
    privileges => {
      'cluster' => 'manage_index_templates',
      'indices' => {
        names      => ['logstash-*', 'myapp-*'],
        privileges => [
          'write',
          'delete',
          'create_index',
        ],
      },
    },
  }
```

Hope this helps, perhaps again it's worth adding an example like this in the docs.

fgimian:
OK one more thing, how do I apply SSL certs specifically to Shield like this?

```
shield:
  ssl:
    keystore:
      password: abc123
      path: /etc/elasticsearch/client-node/shield/bla.jks
```

Thanks so much!!
Fotis

tylerjl:
Thank you for testing @fgimian! Additional documentation added.

PR #627, opened 2016-05-11
Handle Restart/RestartSec systemd service file options

eLvErDe:
Hello,

Here is a quick improvement to be able to use auto-restart feature provided by systemd service files.

Best regards, Adam.

### DIFF
--- a/manifests/service/systemd.pp
+++ b/manifests/service/systemd.pp
@@ -163,6 +163,18 @@ define elasticsearch::service::systemd(
         $memlock = undef
       }
 
+      # Systemd automatic restart feature
+      if ($new_init_defaults != undef and is_hash($new_init_defaults) and has_key($new_init_defaults, 'SYSTEMD_RESTART')) {
+        $systemd_restart = $new_init_defaults['SYSTEMD_RESTART']
+      }else{
+        $systemd_restart = undef
+      }
+      if ($new_init_defaults != undef and is_hash($new_init_defaults) and has_key($new_init_defaults, 'SYSTEMD_RESTART_SEC')) {
+        $systemd_restart_sec = $new_init_defaults['SYSTEMD_RESTART_SEC']
+      }else{
+        $systemd_restart_sec = undef
+      }
+
       file { "${elasticsearch::params::systemd_service_path}/elasticsearch-${name}.service":
         ensure  => $ensure,
         content => template($init_template),
--- a/templates/etc/init.d/elasticsearch.systemd.erb
+++ b/templates/etc/init.d/elasticsearch.systemd.erb
@@ -19,6 +19,8 @@ LimitMEMLOCK=<%= @memlock %>
 <% end %>
 # Shutdown delay in seconds, before process is tried to be killed with KILL (if configured)
 TimeoutStopSec=20
+<% if @systemd_restart %>Restart=<%= @systemd_restart %><% end %>
+<% if @systemd_restart_sec %>RestartSec=<%= @systemd_restart_sec %><% end %>
 
 [Install]
 WantedBy=multi-user.target


tylerjl:
This could definitely be useful for people on systemd-based distributions.

Is there a particular reason the two parameters are stored in the `init_defaults` hash? They may be more easily usable as a top-level parameter on `instance` and they won't do anything as shell variables in the init defaults file AFAIK.

eLvErDe:
Hello Tyler,

No specific reason I just saw it was done that way so I mimiced the existing.
I'm not a Puppet expert, just tried to solve my issue and though it could be useful to everyone!

Regards, Adam.

tylerjl:
Thanks @eLvErDe. I think this only needs a few changes changes to merge it in:

* Make `systemd_restart` and `systemd_restart_sec` top-level parameters, which can make them more easily validated in puppet and so they can given their own pieces of documentation for users.
* In the template, use the erb `-%>` notation to avoid interpolating empty lines  into the systemd unit template. For example, this
```erb
<% if @systemd_restart %>Restart=<%= @systemd_restart %><% end %>
```
Becomes this:
```erb
<% if @systemd_restart -%>Restart=<%= @systemd_restart %><% end -%>
```

eLvErDe:
Hello,

Thanks for the feedback. When you say top level you mean parameter of the elasticsearch class directly?

Regards, Adam

Le 24 mai 2016 22:01:06 GMT+02:00, Tyler Langlois <notifications@github.com> a écrit :
>Thanks @eLvErDe. I think this only needs a few changes changes to merge
>it in:
>
>* Make `systemd_restart` and `systemd_restart_sec` top-level
>parameters, which can make them more easily validated in puppet and so
>they can given their own pieces of documentation for users.
>* In the template, use the erb `-%>` notation to avoid interpolating
>empty lines  into the systemd unit template. For example, this
>```erb
><% if @systemd_restart %>Restart=<%= @systemd_restart %><% end %>
>```
>Becomes this:
>```erb
><% if @systemd_restart -%>Restart=<%= @systemd_restart %><% end -%>
>```
>
>---
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/elastic/puppet-elasticsearch/pull/627#issuecomment-221377333

-- 
Envoyé de mon appareil Android avec K-9 Mail. Veuillez excuser ma brièveté.

tylerjl:
Correct; placing them in `init.pp` means every instance on the host will get the same settings but we can add per-instance parameter overrides later if needed.

`init.pp` would also be a good place to validate that the values assigned to the parameters are valid for systemd (i.e. check that `Restart=` is on-failure, on-success, etc., and `RestartSec=` is an integer).

eLvErDe:
Got it, I'll try to find time to do this tomorrow. Should I fail() if parameters are invalid?

Le 24 mai 2016 23:20:32 GMT+02:00, Tyler Langlois <notifications@github.com> a écrit :
>Correct; placing them in `init.pp` means every instance on the host
>will get the same settings but we can add per-instance parameter
>overrides later if needed.
>
>`init.pp` would also be a good place to validate that the values
>assigned to the parameters are valid for systemd (i.e. check that
>`Restart=` is on-failure, on-success, etc., and `RestartSec=` is an
>integer).
>
>---
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/elastic/puppet-elasticsearch/pull/627#issuecomment-221398018

-- 
Envoyé de mon appareil Android avec K-9 Mail. Veuillez excuser ma brièveté.

tylerjl:
Yep! I'd just follow the way the other parameters are validated (like `status` for example).

fatmcgav:
Hi there,

Thanks again.

TheMeier:
There is an unresolved merge-conflict for 3 yrs+. 
I am closing it just because of the inactivity. **Please** re-open if there is still interest int fixing this.

PR #628, opened 2016-05-11
Update to current apt module syntax

eLvErDe:
Apt module issued deprecation warning, here is an updated call that match current syntax.

Regards, Adam.


tylerjl:
A few notes:

* There's a test failure for updated documentation on the parameter
* Those changed parameters mean that `metadata.json` should be updated to reflect the minimum required version of the `apt` module so people with old versions of the module don't suffer broken puppet runs if they're on old versions of the module and the parameters aren't accepted

tylerjl:
Updated in #858

Issue #336, opened 2015-05-21
restart_on_change should have a different default


faxm0dem:
I think `elasticsearch::restart_on_change = false` should be the default value, as long as there is no [elasticsearch rolling restart API call](/elastic/elasticsearch/issues/9046). Indeed, using puppet to manage multiple cluster nodes could have catastrophic consequences with this parameter set to `true`.

electrical:
I tend to agree with this since automatic restarts on config changes is something we wouldn't want.
Feel free to send a PR otherwise i'll make the change my self.

tylerjl:
Changed all default restart_* parameters to `false` in #655, changes will be part of the next release (1.0.0).

PR #635, opened 2016-05-17
Closes #618

### LINKED ISSUE #618: Config hash can't contain array of hashes
I'm installing an ES plugin via puppet that uses a couple of parameters in the `elasticsearch.yml` file. Some of these parameters are supposed to be made up out of an array of hashes, like this:
```ruby
data => [ { key       => "some value 0",
            other_key => "some other value 0" },
          { key       => "some value 1",
            other_key => "some other value 1" } ]
```
This should result in the following yaml snippet:
```ruby
  data:
    - key: some value 0
      other_key: some other value 0
    - key: some value 1
      other_key: some other value 1
```

However, when I added these parameters to the `config` hash that is passed to the `elasticsearch::instance` define, puppet threw me the following error: "Could not evaluate: can't convert Hash into String". 

After checking the `elasticsearch.yml.erb` template I noticed that the `recursive_hash_to_yml_string` method doesn't check if an array value is a hash, so I changed the method to the following:

```ruby
  # Function to make a structured and sorted yaml representation out of a hash
  def recursive_hash_to_yml_string(hash, depth=0)
    spacer = ""
    depth.times { spacer += "  "}
    hash.keys.sort.each do |sorted_key|
      @yml_string += spacer + sorted_key + ": "
      if hash[sorted_key].is_a?(::Array)

         keyspacer = "  "
         # I can't see why to do this, replaced by keyspacer = "  "
         #sorted_key.length.times { keyspacer += " " }

         @yml_string += "\n"
         hash[sorted_key].each do |item|

            # Is the current item a hash?
            if item.is_a?(::Hash)
               @yml_string += spacer + keyspacer + "- \n"
               recursive_hash_to_yml_string(item, depth+2)
            else
               @yml_string += spacer + keyspacer + "- " + item +"\n"
            end

         end
      elsif hash[sorted_key].is_a?(::Hash)
        @yml_string += "\n"
        recursive_hash_to_yml_string(hash[sorted_key], depth+1)
      else
        @yml_string += "#{hash[sorted_key].to_s}\n"
      end
    end
  end
```

Given the array above, it will now generate the following yaml code:
```ruby
  data:
    -
      key: some value 0
      other_key: some other value 0
    -
      key: some value 1
      other_key: some other value 1
```
This includes an extra newline after the `-` but it seems ES as well as the plugin are perfectly happy with this!

1n:
As part of your pull request, please review the following tasks:

- [x] Tests pass

Rspec tests for ERB templates pass for puppet 3.8.0 and puppet >=4.0.0.

Unfortunately have nothing to do with the rest of YAML implementation differences for Puppet 3 and 4 except this:
```ruby
    # Puppet < 4 uses ZAML, which has some deviations from Puppet 4 YAML implementation
    unless Puppet::Util::Package.versioncmp(Puppet.version, '4') >= 0
      @yml_string.gsub!(/^\s{2}/, '')
      # Remove space character after colon
      @yml_string.gsub!(/:\s$/, ':')
      # Remove space character after ---
      @yml_string.gsub!(/---\s$/, '---')
      # Add new line at the end of yaml file
      @yml_string += "\n"
    end
```
So updated rspec tests to handle this difference. Some of the tests are more demonstrative to show difference. Added YAML validation to rspec tests.

Maybe more tests, validation is required because this config is rather important.

### DIFF
--- a/spec/acceptance/013_config_spec.rb
+++ b/spec/acceptance/013_config_spec.rb
@@ -41,7 +41,7 @@ describe "elasticsearch class:" do
       it { should contain 'master: true' }
       it { should contain 'data: false' }
       it { should contain "cluster:\n  name: #{test_settings['cluster_name']}" }
-      it { should contain 'rack: 46' }
+      its(:content) { should match /rack: \W*46\W*/ }
       it { should contain "index: \n  routing: \n    allocation: \n      exclude: \n             - tag2\n             - tag3\n      include: tag1" }
     end
 
--- a/spec/templates/001_elasticsearch.yml.erb_spec.rb
+++ b/spec/templates/001_elasticsearch.yml.erb_spec.rb
@@ -1,4 +1,5 @@
 require 'spec_helper'
+require 'yaml'
 
 class String
   def config
@@ -8,6 +9,13 @@ class String
   def unindent
     gsub(/^#{scan(/^\s*/).min_by{|l|l.length}}/, "")
   end
+end
+
+# Puppet 4 YAML implementation doesn't indent array elements
+if Puppet.version >= '4.0.0'
+  indent = ''
+else
+  indent = '  '
 end
 
 describe 'elasticsearch.yml.erb' do
@@ -29,19 +37,22 @@ describe 'elasticsearch.yml.erb' do
       }
     )
 
-    expect(harness.run).to eq(%q{
+    expect(harness.run).to eq(%Q{
       discovery:
         zen:
           ping:
             unicast:
               hosts:
-                - host1
-                - host2
+              #{indent}- host1
+              #{indent}- host2
       node:
         name: test
       path:
         data: /mnt/test
       }.config)
+
+    # Check for a valid YAML
+    expect{ YAML.load(harness.run) }.to_not raise_error
   end
 
   it 'should merge hashes' do
@@ -57,6 +68,8 @@ describe 'elasticsearch.yml.erb' do
         name: test
         rack: r1
       }.config)
+
+    expect{ YAML.load(harness.run) }.to_not raise_error
   end
 
   it 'should concatenate arrays' do
@@ -67,18 +80,20 @@ describe 'elasticsearch.yml.erb' do
       }
     )
 
-    expect(harness.run).to eq(%q{
+    expect(harness.run).to eq(%Q{
       data:
         path:
-          - /mnt/sda1
-          - /mnt/sdb1
+        #{indent}- /mnt/sda1
+        #{indent}- /mnt/sdb1
       }.config)
+
+    expect{ YAML.load(harness.run) }.to_not raise_error
   end
 
   it 'should qualify conflicting hash keys' do
     harness.set(
       '@data', {
-        'shield.http.ssl' => 'true',
+        'shield.http.ssl' => true,
         'shield.http.ssl.client.auth' => 'optional'
       }
     )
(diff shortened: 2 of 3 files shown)


tylerjl:
I think the only legitimate failure I caught in the tests was [this one](https://travis-ci.org/elastic/puppet-elasticsearch/jobs/130843088#L1703):
```
Failure/Error: it { should contain 'rack: 46' }
  expected File "/etc/elasticsearch/es-01/elasticsearch.yml" to contain "rack: 46"
```
I'll try and reproduce this as well to see how we may need to munge the yaml string. My guess is either the tests or the yaml is up to something odd and there's quotes or something similar being inserted in yaml.

(By the way, I manually cancelled the tests since we're working through some capacity issues, but you should be able to safely push to trigger more tests now.)

tylerjl:
Tested this against < 4 and >= 4 versions of puppet, the main differences seem to be regarding quoted versus unquoted values. For example, >= 4 `to_yaml` places quotes around file paths, earlier versions don't.

I tested how Elasticsearch reacts to quoted numerics, and it parses them just fine (for example, using `http.port: "9201"` will cause ES to bind on port 9201). Given that the differences in yaml are kind of quirky but end results are the same, we may have to rely on re-parsing the template content to test for equivalency rather than character-for-character string comparisons.

1n:
@tylerjl do you mean here:

> we may have to rely on re-parsing the template content to test for equivalency rather than character-for-character string comparisons.

something like this?
```ruby
it 'should merge hashes' do
    harness.set(
      '@data', {
        'node.name' => 'test',
        'node.rack' => 'r1'
      }
    )

    expect( YAML.load(harness.run) ).to eq(YAML.load(%q{
      node:
        name: test
        rack: r1
      }.config))
end
```
Regarding 
> Failure/Error: it { should contain 'rack: 46' }
  expected File "/etc/elasticsearch/es-01/elasticsearch.yml" to contain "rack: 46"

it could be
```ruby
    describe file('/etc/elasticsearch/es-01/elasticsearch.yml') do
      it { should be_file }
      it { should contain 'name: elasticsearch001' }
      it { should contain 'master: true' }
      it { should contain 'data: false' }
      it { should contain "cluster:\n  name: #{test_settings['cluster_name']}" }
      its(:content) { should match /rack: \W*46\W*/ }
      it { should contain "index: \n  routing: \n    allocation: \n      exclude: \n             - tag2\n             - tag3\n      include: tag1" }
    end
```
But there could be some more places.

Does acceptance tests run also for Puppet >= 4.0.0?

tylerjl:
Yep, that's what I was thinking. It won't test the contents explicitly but it's probably good enough (I mean, the templates haven't even had unit tests until the last couple of days, so this is still definitely an improvement.)

Totally on-board with changing the `contents` check like you did in the commit.

There aren't any acceptance tests are explicitly test against Puppet >= 4, but all Puppet Enterprise tests for version 2016.1.1 for example use Puppet 4.x under the hood, so 4.x gets tested anyway. 👍

tylerjl:
Based upon the test failures, I'm pretty sure the idempotency tests failed, indicating the hash is getting re-written in a different order (it looks like it only got caught on the old distros/those with Ruby <= 1.8). We may need to use something like the to_yaml hash ordering gymnastics I did in #638 if we want to make to_yaml consistent.

1n:
Looks like tests are ok, except for a failed puppet-enterprise installation on SLES.

tylerjl:
Thanks for all your work on this, @1n. That failure is a transient package issue I'm tracking for for SLES 12, but everything else looks great. The only thing I can possibly think of this may effect would be some yaml values getting quoted with the change in `to_yaml`, but I think we'll add in #336 to the next release to avoid unexpected service restarts which should mitigate those sort of changes.

Cheers!

PR #639, opened 2016-05-25
Removed install_options

AshtonDavis:
install_options was causing all sorts of issues.  In lieu of fixing it, right now I've opted to remove it.  The behavior exhibited by the install_options parameter was inconsistent at best and upon looking at the (few) lines of code involving install_options, I couldn't determine a reason for it to sometimes report as an invalid parameter.

### DIFF
--- a/lib/puppet/provider/elasticsearch_plugin/plugin.rb
+++ b/lib/puppet/provider/elasticsearch_plugin/plugin.rb
@@ -86,19 +86,11 @@ Puppet::Type.type(:elasticsearch_plugin).provide(:plugin) do
     commands
   end
 
-  def install_options
-    return @resource[:install_options].join(' ') if @resource[:install_options].is_a?(Array)
-    return @resource[:install_options]
-  end
-
   def create
     es_version
     commands = []
     commands << @resource[:proxy_args].split(' ') if @resource[:proxy_args]
-    commands << install_options if @resource[:install_options]
-    if install_options.nil? or not install_options.include? 'es.path.conf'
-      commands << "-Des.path.conf=#{homedir}"
-    end
+    commands << "-Des.path.conf=#{homedir}"
     commands << 'install'
     commands << '--batch' if is22x?
     commands << install1x if is1x?
--- a/lib/puppet/type/elasticsearch_plugin.rb
+++ b/lib/puppet/type/elasticsearch_plugin.rb
@@ -28,8 +28,4 @@ Puppet::Type.newtype(:elasticsearch_plugin) do
     defaultto '/usr/share/elasticsearch/plugins'
   end
 
-  newparam(:install_options) do
-    desc 'Installation options'
-  end
-
 end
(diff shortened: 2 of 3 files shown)


AshtonDavis:
I have signed the CLA.

lib/puppet/provider/elasticsearch_plugin/plugin.rb
```diff
@@ -86,19 +86,10 @@ def install2x
...
-    return @resource[:install_options]
-  end
-
   def create
     es_version
     commands = []
     commands << @resource[:proxy_args].split(' ') if @resource[:proxy_args]
-    commands << install_options if @resource[:install_options]
-    if install_options.nil? or not install_options.include? 'es.path.conf'
-      commands << "-Des.path.conf=#{homedir}"
```
tylerjl: This particular java option needs to stay in order to make plugins like Shield install their configurations into `/usr/share/elasticsearch` rather than `/etc/elasticsearch`

tylerjl:
Hey @AshtonDavis,

I made one note on the commit regarding a line that we need for Shield functionality.

Looking at the flags for `bin/plugin`, I'm inclined to agree that there actually aren't many (if any) options that would be relevant for config management to set when installing the plugins; I'll look around some more to ensure this won't terribly break anything.

Out of curiosity, what problems is `install_options` causing in your case? Is it something we could fix in-place rather than purge in case someone is relying on `install_options`?

AshtonDavis:
I didn't realize electrical wasn't working at elastic anymore - I submitted this PR after a conversation with him.  I'll fix and resubmit my commit in a second, thanks for pointing out that line is needed.

At least once every few hours (sometimes many runs in a row) I will get an invalid parameter error regarding install_options, and there wasn't anything I found that would fix it.  The install options, according to electrical, was intended to allow for staging installs.  I don't believe he documented it or made it available as a feature anywhere, so I doubt it affects anyone if it were removed.

Based on the logic in the three files containing install_options I couldn't see anything immediately obvious.  Here's what the error looks like:
```
Could not retrieve catalog from remote server: Error 400 on SERVER: Invalid parameter install_options on Elasticsearch_plugin[lmenezes/elasticsearch-kopf] at /etc/puppet/environments/development/modules/elasticsearch/manifests/plugin.pp:142 on node elk-01-dev
```

kopf is the only plugin being installed.  Presumably it would complain once for each plugin if I had more than one.

lib/puppet/provider/elasticsearch_plugin/plugin.rb
AshtonDavis: Re-added. :)

tylerjl:
Agreed on all points. Without any documentation and the trouble it seems to cause people fairly often, seems like the best option.

I'll make specific mention of it in the release notes so people are aware.

AshtonDavis:
Fantastic - thanks a lot!

PR #644, opened 2016-06-10
Mask default unit (from package)

sathieu:


sathieu:
This is the most portable way to do it (i.e, no dependency on recent puppet)

tylerjl:
Thanks @sathieu, this does feel a lot more correct than just deleting the file that the package manager drops onto the filesystem. Agree with making this change.

I do have one note, should we get rid of the [file resource that deletes the original unit](https://github.com/sathieu/puppet-elasticsearch/blob/af2f1ace7660f15e2c741ef64a374e0003656c05/manifests/config.pp#L122)? The masking should take care of unloading the unit and so ensuring absent is just redundant now.

sathieu:
it looks like travis is somewhat broken...

tylerjl:
@sathieu it's actually just **very** slow at the moment and is building other PRs/branches, so give it some time (maybe even over a day) to get through the backlog and it should reach your commit.

Sorry about the delay - speeding up the builds and making Travis more efficient for this repo is something I'm looking into. In the meantime just let it do its thing, I'll be sure and look at the test results once they're done. 👍

sathieu:
It is failing on DISTRO=ubuntu-server-1404-x64 PE=1, but probably not because of the patch (SSH connection error)

sathieu:
@tylerjl Do you know why it is failing?

tylerjl:
@sathieu should be a transient failure, I kicked off a re-test.

I think at this point things should be good to merge, I may add a test later to confirm that the unit is symlinked (essentially a replacement for the removed rspec line). But overall I'm 💯  for the change!

I'll merge it after the remaining test passes (or I can wait if you want to add the symlink rspec test yourself, either way).

sathieu:
test paseed. Great!

tylerjl:
Nice, looks good.

I just realized one more thing here: the `file` resource for the link to `/dev/null` is nested within a conditional about the `pid_dir` variable. It should technically be in the same spot as the `file` resource that used to have `ensure => absent`, right after the resource that gets rid of `/etc/init.d/elasticsearch`. It should just be a copy/paste (with a check for the systemd dir).

While you're making that small change, could you replace the deleted rspec check with a check for the symlink? You can probably use the [serverspec `be_linked_to` resource](http://serverspec.org/resource_types.html#file) to make it easier.

Sorry about adding another to-do item, I didn't notice it until now. But they're small changes, this will definitely make it into the next release 👍

PR #645, opened 2016-06-10
Ensure elasticsearch.yml ends with a newline

sathieu:


tylerjl:
Thanks @sathieu, I noticed this when we replaced the template with `to_yaml` but totally forgot to fix the yaml output. Appreciate the work!

Issue #643, opened 2016-06-10
Plugin install fails with "No such file..."


peske:
I've tried installing [dejaVu](https://github.com/appbaseio/dejaVu) plugin by using `elasticsearch::plugin` class, but it has ended with the error:

```
No such file or directory @ rb_sysopen - /usr/share/elasticsearch/plugins/dejavu/.name
```

At first I've thought that it is about the plugin itself, and wanted to report the problem there, but when I've tried to install the plugin directly, without puppet, it was successful:

```
Installed dejaVu into /usr/share/elasticsearch/plugins/dejaVu
```

So obviously the problem is with puppet-elasticsearch module. Is it a bug? Or I'm doing something wrong?

Thanks!

tylerjl:
Interesting issue! Thanks for reporting.

I actually think I know what the problem is - when puppet manages the plugin, it creates the `.name` file to track the plugin metadata for puppet to use later. Based upon what I'm seeing, it may actually be trying to read it in the `dejavu` directory (note the case) but the plugin is installed within a directory with the fancy `dejaVu` case you're seeing.

Do you have information regarding the version of the module you're using, version of puppet, OS and version, and the puppet plugin resource in the manifest you're using to install dejaVu to see if I can reproduce it?

peske:
You've probably right that it is about the case. Regarding my environment: pretty much newest in everything:

- ElasticSearch 2.3 (or 2.3.3? the latest definitely since I've installed it yesterday)
- ElasticSearch Module 0.11.0
- Puppet Agent 4.5.1
- Puppet Master 4.5.1
- Both ElasticSearch and Puppet server are running on Ubuntu 14.04.4
- dejaVu Module - the latest again

My manifest:

```
elasticsearch::plugin { 'appbaseio/dejaVu':
  instances  => $node_name,
  proxy_host => 'internal-proxy.itenlight.com',
  proxy_port => 8080, 
}
```

`$node_name` is `IT-ES01`. I've installed several more plugins without problem:

```
elasticsearch::plugin { 'lmenezes/elasticsearch-kopf':
  instances  => $node_name,
  proxy_host => 'internal-proxy.itenlight.com',
  proxy_port => 8080, 
}

elasticsearch::plugin { 'mobz/elasticsearch-head':
  instances  => $node_name,
  proxy_host => 'internal-proxy.itenlight.com',
  proxy_port => 8080, 
}
```

bisscuitt:
[...]

PR #646, opened 2016-06-13
Do not force downcase on plugin directory names

tylerjl:
As raised in #643, some plugins are case-sensitive and thus create case-sensitive directories in the plugin directory. This PR 1) adds tests for the provider's pluginfile behavior and 2) removes the `downcase` call.

@electrical, if you've got a minute, could you maybe shed some light on why that was put there in the first place? I don't want to introduce a regression since it seems this must have been put here for a specific reason per the line I'm proposing to remove that checks for a 2.x version.


electrical:
Bit late to the party. but main reason was that with discovery of the directory it required the downcased name since most people used lower case in the plugin properties and install required mixed case ( github plugins for example )

PR #655, opened 2016-06-30
Add restart parameters and change default behavior

tylerjl:

- [x] Tests pass

---

This change:

* Changes the default for `restart_on_change` to `false` per and #336
* Adds some additional parameters for additional control per #363 
* Updates all tests to reflect these changes
* Various cleanup items as I came across them


manifests/init.pp
```diff
@@ -230,52 +267,55 @@
...
-  $plugins               = undef,
-  $plugins_hiera_merge   = false,
-  $use_ssl               = false,
-  $validate_ssl          = true,
-  $ssl_user              = undef,
-  $ssl_password          = undef
+  $ensure                 = $elasticsearch::params::ensure,
+  $status                 = $elasticsearch::params::status,
+  $restart_on_change      = $elasticsearch::params::restart_on_change,
+  $restart_config_change  = undef,
```
electrical: Why not have these default to the same restart_on_change value? then you don't need all the rubbish logic below.

tylerjl: I didn't realize I could self-reference parameters here, that's a much better idea - thanks for the suggestion!

electrical: should be able to do `$elasticsearch::restart_on_change`

PR #658, opened 2016-07-12
Fix deprecated include_src

b4ldr:
- [ X ] Signed the [CLA](https://www.elastic.co/contributor-agreement/) (if not already signed)
- [ X ] Rebased/up-to-date with base branch
- [ X ] Tests pass
- [ N/A ] Updated CHANGELOG.md with patch notes (if necessary)
- [ N/A ] Updated CONTRIBUTORS (if attribution is requested)


tylerjl:
Hi @b4ldr, thanks for the contribution. Note that the `include` parameter was added only as of version 2.x, so you also need to update `metadata.json` to reflect that the new minimum required version of the `puppetlabs/apt` module is version 2 or greater, otherwise this module has the potential to break if any apt module version older than 2 gets installed as a module dependency.

PR #660, opened 2016-07-15
Allow override of module location

bisscuitt:
This is an attempt to fix / work around issue #643 by allowing an override of the install location for a plugin.

I don't like the way I'm re-using a previously deprecated parameter and I'm not happy with the resource naming I'm using... but it atleast works as a proof of concept.

Sucessfull puppet run using `swift-repository` from #643  looks like:
```
Notice: /Stage[main]/Bigcommerce_elasticsearch::Plugins/Bigcommerce_elasticsearch::Plugin[org.wikimedia.elasticsearch.swift/swift-repository-plugin/2.3.3.1]/Elasticsearch::Plugin[org.wikimedia.elasticsearch.swift/swift-repository-plugin/2.3.3.1]/Elasticsearch_plugin[org.wikimedia.elasticsearch.swift/swift-repository-plugin/2.3.3.1]/ensure: created
```

Tests probably need updating to reflect the changes too.

- [ ] Tests pass


bisscuitt:
Looks like test failure is unrelated and is down to a docker issue.
```
Provisioning docker
provisioning sles-12-x64
Using docker server at 0.0.0.0
  Warning: Try 1 -- Host 0.0.0.0 unreachable: Net::SSH::Disconnect - connection closed by remote host
  Warning: Trying again in 3 seconds
Warning: ssh connection to sles-12-x64 has been terminated
```

This seems to have been cropping up on previous test runs on other PRs too. Perhaps this test can be re-ran somehow ?

tylerjl:
I kicked off the test again; some of the tests can be flaky at times. 🌵 

Agreed that this parameter should probably exist to help resolve issues like the one you've highlighted. I also agree that the naming isn't ideal, as it took some reading to figure out what refers to what; so making the naming clearer would probably help. Maybe rename the type's parameter to `module_dir` as well? It's consistent with `plugin.pp` then, although it's still not 100% accurate.

tylerjl:
Alternatively, we could make the `pluginfile` method smarter to attempt to auto-detect the directory name by looking for matching directory names after stripping `-plugin` off the directory names. Are there any other cases you've seen where this behavior exists? I'm trying to figure out if this is a pattern with plugins (adding `-plugin` to the name but it disappearing upon installation.)

bisscuitt:
@tylerjl I've not seen this with other plugins. I thinkn trying to find based on munging the plugin name could be pretty flakey. I tihnk the point is that plugins may create a directory different to what the plugin is called. So having a mechanism to override the directory name would proably be the best bet, rather than trying to make the module guess based on a practically infinte number of random possibilities.

I'd appreciate some more direction on naming of the parameters and variables.

tylerjl:
You're probably right; the auto-detection could really only get us so far. Let's stick with exposing a parameter.

As far as naming, I think we'll probably just have to live with it for now - please do comment on the commit if anyone can come up with a better naming scheme, at that point we can deprecate it and eventually purge it when we do a major point release. 👍 

The tests look fine to me; SLES is an ever-present source of pain. Do you want to add yourself to CONTRIBUTORS.md or are you ready for me to merge? (I can add the changelog notes if you'd like).

bisscuitt:
I've updated `CONTRIBUTORS`. Would appreciate if if you could add the changelog notes.

Issue #152, opened 2014-07-24
Allow installation of plugins from behind a proxy


FergusNelson:
The elasticsearch plugin command does not use environment proxy settings for retrieving the plugins. However it does support command line args for proxy settings.  

/usr/share/elasticsearch/bin/plugin -DproxyPort=$proxy_port -DproxyHost=$proxy_host 

The plugin resource should expose these parameters explicitly. As a work around i have overridden the $plugintool parameter 

$plugintool = "/usr/share/elasticsearch/bin/plugin -DproxyPort=$proxy_port -DproxyHost=$proxy_host "

muelli:
It doesn't seem to work for me.
How is that supposed to work for proxies requiring authentication?  What values should I put for `proxy_host` and `proxy_port`?

electrical:
@muelli i think you could do something like this for the auth part: `proxy_host = 'user:pass@host'`

muelli:
hm. that doesn't seem to work for me :-(
Also, https://github.com/elastic/elasticsearch/commit/cf7ebfcebf50ad596bf5d6f373da94f01500daf9 mentions `-Dhttp.proxyHost`, not `-Dproxy_host`. But it remains to be seem if there is a working parameter for authenticated proxies. So far, I haven't managed to get it working.

Why can the plugin installer not simply take the environment variable? I've filed a bug about it: https://github.com/elastic/elasticsearch/issues/11001

joanmarcriera:
Same problem here .

muelli:
the problem is that the elastic people seem to be unwilling to respect that GNU/Linux hosts are indeed a platform that people run their ES instances on. At least they deem it not important enough to support the proxy use case as can be seen in the other bug report https://github.com/elastic/elasticsearch/issues/16633 which might also contain a workaround.

tylerjl:
[...]

PR #670, opened 2016-07-29
Support username/password for plugin installations behind proxies

tylerjl:

- [x] Tests pass

---

This should support all the parameters mentioned in #152. I opted to pass a proxy URL through to the provider to avoid potentially crazy regex parsing and just let `URI()` handle parsing the relevant parts of the proxy URL instead.

There's some general cleanup sprinkled in here as well (removed an unused parameter, whitespace, deduplicated plugin tests).


manifests/plugin.pp
```diff
@@ -48,6 +48,18 @@
...
 #   This variable is optional
 #
+# [*proxy_username*]
+#   Proxy auth username to use when installing the plugin
+#   Value type is string
+#   Default value: None
+#   This variable is optional
+#
+# [*proxy_password*]
+#   Proxy auth username to use when installing the plugin
```
JGroselle: Too fast copy/paste. I'm sure that youwanted to write password instead.
's/username/password/'

tylerjl: Whoops, good catch, thanks! 👍

Issue #688, opened 2016-08-24
Failing to apply elasticsearch::template


pabrahamsson:
* Module version: `0.13.1`
* Puppet version: `3.8.7`
* OS and version: `Ubuntu 14.04`

## Bug description

Error message
```
$ sudo puppet agent -t
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
Info: Caching catalog for marvel01.example.com
Info: Applying configuration version ''
Error: Could not prefetch elasticsearch_template provider 'ruby': end of file reached
Notice: /Stage[main]/Role::Marvel/Elasticsearch::Template[marvel]/Elasticsearch_template[marvel]/ensure: created
Error: /Stage[main]/Role::Marvel/Elasticsearch::Template[marvel]/Elasticsearch_template[marvel]: Could not evaluate: end of file reached
Notice: Finished catalog run in 13.38 seconds
```

Manifest
```
elasticsearch::template { 'marvel':
  content => 
    'order' => 1,
    'settings' => {
      'number_of_replicas' => 0
    },
    'template' => ".marvel*"
  }
}
```

Hiera
```
elasticsearch::api_basic_auth_password: <eyaml encrypted password>
elasticsearch::api_basic_auth_username: admin
elasticsearch::api_host: "%{::fqdn}"
elasticsearch::api_protocol: https
```

Some additional info
```
/var/lib/puppet/lib/puppet/provider/elasticsearch_template$ ruby ruby.rb
/usr/lib/ruby/1.9.1/rubygems/custom_require.rb:36:in `require': cannot load such file -- puppet_x/elastic/deep_to_i (LoadError)
           from /usr/lib/ruby/1.9.1/rubygems/custom_require.rb:36:in `require'
           from ruby.rb:7:in `<main>'
/var/lib/puppet/lib/puppet/provider/elasticsearch_template$ cd ../../..
/var/lib/puppet/lib$ ruby puppet/provider/elasticsearch_template/ruby.rb
/usr/lib/ruby/1.9.1/rubygems/custom_require.rb:36:in `require': cannot load such file -- puppet_x/elastic/deep_to_i (LoadError)
           from /usr/lib/ruby/1.9.1/rubygems/custom_require.rb:36:in `require'
           from puppet/provider/elasticsearch_template/ruby.rb:7:in `<main>'
/var/lib/puppet/lib$ ll puppet_x/elastic/
total 20
[...]

PR #689, opened 2016-08-25
Fix elasticsearch::template HTTPS usage

tylerjl:

- [x] Tests pass

Per #688, the ruby provider actually never sets the SSL flag for the Net::HTTP object it uses (whoops!) This change sets the flag appropriately and also adds an explicit test to ensure we're validating it does so.


pabrahamsson:
Thank you @tylerjl, just wanted to let you know that the template was successfully created once I applied this PR.

PR #727, opened 2016-10-28
Update the apt::source call to not cause deprecation warnings

sgaduuw:
- [v] Signed the [CLA](https://www.elastic.co/contributor-agreement/) (if not already signed)
- [v] Rebased/up-to-date with base branch
- [ ] Tests pass
- [v] Updated CHANGELOG.md with patch notes (if necessary)

Update the apt::source call to not cause deprecation warnings
[puppetserver] Scope(Apt::Source[elasticsearch]) $include_src is deprecated and will be removed in the next major release, please use $include => { 'src' => false } instead
[puppetserver] Scope(Apt::Source[elasticsearch]) $key_source is deprecated and will be removed in the next major release, please use $key => { 'source' => http://packages.elastic.co/GPG-KEY-elasticsearch } instead.

Gave up trying to test with provided tooling since they don't work and trying to get them to work kills the benefits of a 2-minute fix to an apt::source call. Code is being used in production, and works as intended, you'll have to take my word for it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+## x.x.x (Month Day, Year)
+
+### Summary
+
+#### Features
+
+#### Bugfixes
+
+#### Changes
+
+#### Testing changes
+
 ## 0.14.0 (October 12, 2016)
 
 ### Summary
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Features
 
 #### Bugfixes
+* Update the apt::source call to not cause deprecation warnings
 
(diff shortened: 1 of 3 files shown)


tylerjl:
Hi @sgaduuw , thanks for the contribution. Note that the include parameter was added only as of version 2.x, so you also need to update metadata.json to reflect that the new minimum required version of the puppetlabs/apt module is version 2 or greater, otherwise this module has the potential to break if any apt module version older than 2 gets installed as a module dependency.

sgaduuw:
Hi @tylerjl , metadata.json has been updated to reflect the new version requirement.

PR #730, opened 2016-11-11
allow contolling the boot sequence order for systemd

schwicke:

- [ ] Signed the [CLA](https://www.elastic.co/contributor-agreement/) (if not already signed)
- [ x ] Rebased/up-to-date with base branch
- [ ] Tests pass
- [ x ] Updated CHANGELOG.md with patch notes (if necessary)


schwicke:
Use case for this merge request: 
we're using autofs to mount an NFS volume on our ES cluster for backups. If one node is rebooted, it can happen that ES is started before autofs so that ES fails because the backup volume is not available. This can be avoided by ensuring that ES is started after autofs with this patch.

files/etc/init.d/elasticsearch.systemd.erb
```diff
@@ -1,8 +1,15 @@
+<%
```
tylerjl: This erb interpolation should be encased in `<%- code -%>` rather than `<% code %>` in order to avoid empty newlines in the unit file.

files/etc/init.d/elasticsearch.systemd.erb
```diff
@@ -1,8 +1,15 @@
...
+order = ["network-online.target"]
+if @resource[:service_order]
+   order.push(@resource[:service_order])
+end
+%>
+
```
tylerjl: Newline should be removed here.

CHANGELOG.md
```diff
@@ -1,20 +1,19 @@
-## x.x.x (Month Day, Year)
+## 0.15.1 (November 11, 2016)
```
tylerjl: Could you leave the existing changelog lines as-is and just add your changes to the relevant Summary, Features, etc. sections? The version, release date, and existing entries in the changelog should be the same as they were.

tylerjl:
Thanks for the updated changeset, @schwicke. The only part left before we can accept the PR is to sign the [Contributor License Agreement](https://www.elastic.co/contributor-agreement).

schwicke:
Hi, sorry for the delay in signing the agreement. It requires some bureaucracy on our side.

tylerjl:
No worries @schwicke - we're just as much contributing to the need for beauracracy here 😉

schwicke:
@tylerjl I finally got some news about the contributor license agreement from our legal department. I will send you a message via linked-in with their requests.

PR #733, opened 2016-11-17
OpenBSD doesn't have sysctl vm.max_map_count, so limit its usage

buzzdeee:

- [x] Tests pass

OpenBSD doesn't have sysctl vm.max_map_count, so limit its usage
to Linux kernels.

This prevents warning messages for me, since on OpenBSD, this sysctl is not there, and looking at metadata.json, there are only Linux systems listed, so I decided to limit that sysctl to the ::kernel 
fact that it is only issued on Linux systems.

Let me know if that is fine, or if other tweaks/enhancements to it are required.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -126,8 +126,11 @@ class elasticsearch::config {
       }
     }
 
-    sysctl { 'vm.max_map_count':
-      value => 262144,
+    # Other OS than Linux may not have that sysctl
+    if $::kernel == 'Linux' {
+      sysctl { 'vm.max_map_count':
+        value => 262144,
+      }
     }
 
   } elsif ( $elasticsearch::ensure == 'absent' ) {


tylerjl:
Thanks a ton @buzzdeee, OpenBSD totally slipped my mind. This should be a safe change and looks totally correct.

Out of curiosity (given that you may have already tested 5.x on BSD), do any of the bootstrap checks fail as they do on Linux for these types of checks?

buzzdeee:
@tylerjl I'm not sure what you are talking about, what bootstrap checks?

tylerjl:
When starting 5.x, it performs a number of checks (like looking at max_map_count) and actually refuses to start unless they're set to the correct values (things like file descriptor limits also go through the same checks). The max_map_count change and some others fixes this for Linux but I don't know if there are similar checks for *BSD.

buzzdeee:
ah, here is the README of the ES package:
http://cvsweb.openbsd.org/cgi-bin/cvsweb/~checkout~/ports/textproc/elasticsearch/pkg/README?rev=1.4&content-type=text/plain

I have bumped my kern.maxfiles to 32768, and my /etc/login.conf contains:

elasticsearch_es01:\
        :openfiles=12000:\
        :tc=daemon:

I take care to set/provide those values outside the ES Puppet module, since I haven't yet dared to get the module to take care of if ;)

with these, no problem to startup ES. But I have to admit, it's only a single node instance, on my private home network, collecting logs, puppet reports and netflow stuff from about 5 - 10 nodes. Indices older than 30 days are deleted by a nightly job, indices older than a day are optimized. So it's not an overly huge/impressive installation ;)

tylerjl:
No problem, I appreciate the info as this can serve as history if we need to get the module to automatically handle it for *BSD users. As it currently stands, getting 5.x support out with just the basics is a good place to start, so we can revisit automatically handling it if we need to later.

PR #755, opened 2016-12-23
Deprecated as per https://github.com/puppetlabs/puppetlabs-stdlib

considerable:

- [x] Tests pass


considerable:
No comments?

tylerjl:
Looking at the stdlib notes for `validate_legacy`, the README indicates that it's only supported in Puppet >= 4.4.0, is that correct? If so, we'd probably want to wait on this change until we drop support for all versions of Puppet older than 4.4.0.

tylerjl:
Sorry for the delayed response, holiday and the accompanying holiday task backlog builds up. 😅

considerable:
I see. The doc https://github.com/puppetlabs/puppetlabs-stdlib#validate_legacy
Says "Note: This function is compatible only with Puppet 4.4.0 (PE 2016.1) and later."
Yet my is puppet 4.7.0 sends a ton of warning on functions like is_bool, is_array and friends.
Are you sure you don't want me to refactor it as following?

if ( versioncmp( $::clientversion, '4.4.0' ) =>0 )

tylerjl:
I see what you're saying - if you'd like to put in the effort to wrap the checks in the conditional block you provided, then I'd be totally fine with that. My main concern was breaking compatibility with users on older version of puppet (which tends to be most people, unfortunately) but using that check should keep it backwards-compatible and also save you from drowning in deprecation notices.

One more thing to note is that `validate_legacy` is a newer function (from what I can tell) so you should also update `metadata.json` to reflect the new minimum required version of `stdlib` to ensure people don't get unknown function errors if they're using older versions of stdlib.

Needless to say, your contributions here are very appreciated. 🎊

considerable:
I see metadata.json already updated. So I just updated conditions like 

if ( versioncmp( $::clientversion, '4.4.0' ) >= 0 )

tylerjl:
@considerable note that the changes you saw in `metadata.json` are unrelated to stdlib, you probably still need to find the new minimum version required to update it.

considerable:
@tylerjl - thanks, you are right. I just updated it.
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.6.0 < 5.0.0"
+      "version_requirement": ">= 4.13.1 < 5.0.0"
     },

tylerjl:
Thanks again for the work @considerable. I noticed that our CI doesn't actually test against any versions of Puppet 4.x (whoops!) so I've retooled some of the Travis testing to more fully test across different versions. If you could get the latest master into your PR, Travis should test both Puppet 3.x and 4.x on some different distributions to ensure we aren't missing anything (sorry for the delay, but thankfully your worked kicked me into getting our test matrix more complete, so thank you!)

tylerjl:
An update for you @considerable: I've found that one of the libraries the module uses to install modules for tests won't properly update the module version installed if the version is changed - I branched off your branch to come up with a fix, I'll PR my fix into your branch once I've figured out the problem. Thanks for your patience, sorry there's been so many weird issues come up with this fix. 🐛

karmi:
Hi @considerable, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/puppet-elasticsearch/pull/755.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

karmi:
Hi @considerable, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/puppet-elasticsearch/pull/755.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

tylerjl:
This is an old one; but the move to Puppet 4.x alleviated the need for `validate_*` calls. 👍

PR #757, opened 2016-12-27
Fix connect timeout

MemberIT:
Fixes connection timeout in es_facts if ES instance has trouble.


tylerjl:
Thank you for the contribution @MemberIT! The change looks good to me, so signing the CLA is the only remaining requirement to merge here.

Issue #734, opened 2016-11-18
installation fails when plugin already existing


tobiasschaber:
Hi,

if I install elasticsearch via puppet and hiera, and add this to my hiera:

elasticsearch::plugins:
  x-pack:
    instances:
      - es-01

re-provisioning fails with the following result:

==> elkmaster1: ERROR: plugin directory /usr/share/elasticsearch/plugins/x-pack already exists. To update the plugin, uninstall it first using 'remove x-pack' command> at 168:/etc/puppet/modules/elasticsearch/manifests/plugin.pp

Maybe I'm doing something wrong but it looks like idempotence is not guaranteed here.

Kind regards,
Tobias

tylerjl:
I @tobiasschaber, can you report:

- The module version this occurs on
- The Linux distro and version
- What version of Elasticsearch (and/or the `class { 'elasticsearch' :` part)
- Puppet version

MoisesRT:
Hi @tylerjl, I am having the same issue.
here is my report:
The error I am receiving after first instalment is
> Invalid parameter proxy on Elasticsearch_plugin[discovery-ec2]
Module: discovery-ec2
Linux: Ubuntu 14.04.3 LTS
ES: 5.0.1
Puppet: 3.4.3 on the node and 3.8.4 on the master

tobiasschaber:
Hey @tylerjl,
I am using:
module version: 0.15.0
linux: centos-6.7
elasticsearch 5.0.1
puppet: 3.8.7-1

This is my class part:

    class { 'elasticsearch' :
        ensure => present,
        status => enabled,
    }

Kind regards,
Tobias

tylerjl:
Have you tried following the error message to remove the plugin before re-running puppet?

That error message appears because puppet drops additional metadata into the plugin directory when it manages plugins for you in order to better guess plugin versions, and those bits of metadata aren't present in plugins you installed by hand so puppet tries to install the plugin regardless. Removing all manually installed plugins and letting puppet handle them should create the metadata and make any future runs idempotent.

MoisesRT:
@tylerjl 
I didn't installed the plugin manually,
I get the following message at the second puppet run:

> Error: /Stage[main]/Logzio-elasticsearch/Elasticsearch::Plugin[discovery-ec2]/Elasticsearch_plugin[discovery-ec2]: Could not evaluate: undefined method `scan' for nil:NilClass

Thanks

tylerjl:
[...]

PR #754, opened 2016-12-20
Add java_home parameter for correct plugin execution #619

lucabelluccini:
Pull request acceptance prerequisites:

- [ ] Signed the [CLA](https://www.elastic.co/contributor-agreement/) (if not already signed)
- [ ] Rebased/up-to-date with base branch
- [ ] Tests pass
- [ ] Updated CHANGELOG.md with patch notes (if necessary)
- [ ] Updated CONTRIBUTORS (if attribution is requested)

Hello, this is a contribution to being able to solve #619.

It is not the best solution, since I wanted somehow pass ALL the env variables, not just the JAVA_HOME (e.g. init_defaults hashmap).

But things start to become complex as init_defaults are merged for hiera support and a plugin is installed per node (and potentially, a different init_defaults per instance).

This is the simpler solution to let anyone using puppet agents with NO JAVA_HOME defined to run and install plugins.

I am proceeding asking for the CLA to my company.

Please review and suggest other fixes/changes on that.

PR #763, opened 2017-01-04
Replace elasticsearch_plugin es_version calls with native calls

tylerjl:

- [ ] Tests pass

Historically the call to `elasticsearch -version` has caused lots of problems due to various underlying causes (`JAVA_HOME`, resource constraints, etc.). This change replaces it with code that calls the `package[elasticsearch]` Puppet catalog resource to retrieve the package version instead.

Related to #734


lucabelluccini:
Does https://github.com/elastic/puppet-elasticsearch/pull/754 make still sense?

tylerjl:
@lucabelluccini it does - although this PR will eliminate the need to call out to ES in `es_version`, all of the commands the provider calls to do things like install/uninstall will still benefit by having `JAVA_HOME` managed in the `with_environment` method (so #754 may need to be updated with this change, but it'll just remove a few lines from the PR).

flysen:
I just stumble on this now. I have a masternode machine with total 16Gb RAM and gave Elastic  `-Xms8g -Xmx8g`. Everything works fine until I do a change and puppet rerun, it runs out of memory.

`Error: /Stage[main]/Main/Node[uuc-logmn012-qa.its.uu.se]/Elasticsearch::Plugin[x-pack]/Elasticsearch_plugin[x-pack]: Could not evaluate: undefined method `scan' for nil:NilClass`

If I log in to the machine and run `/usr/share/elasticsearch/bin/elasticsearch --version`
```
Java HotSpot(TM) 64-Bit Server VM warning: INFO: os::commit_memory(0x00000005d4cc0000, 8241020928, 0) failed; error='Cannot allocate memory' (errno=12)
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (mmap) failed to map 8241020928 bytes for committing reserved memory.
# An error report file with more information is saved as:
# /root/hs_err_pid3438.log
```
```
free -mh
              total        used        free      shared  buff/cache   available
Mem:            15G        8.8G        4.4G         96M        2.3G        6.1G
Swap:            0B          0B          0B
```
If I add an extra 5Gb RAM it works again and puppet works.
So is there an easy way turning of the "duplicate elasticsearch run" in puppet?

tylerjl:
@flysen thanks for confirming some suspicions about this. The call to `elasticsearch` in the plugin provider to determine the version will go away once this PR makes it into a release. As far as I know, that should be the only place the `elasticsearch` executable is directly called by the provider, so it should solve that specific problem - the provider only calls `bin/plugin` to perform plugin operations so hopefully it doesn't run into memory allocation problem.

PR #828, opened 2017-05-15
Adds management for elasticsearch.in.sh file

rhoml:
This PR adds support for managing the elasticsearch.in.sh script. Sometimes as operators we'd like to change parameters on this files for example to enable G1GC or any other task

- [x] Tests pass

### DIFF
--- a/files/usr/share/elasticsearch/bin/elasticsearch.in.sh
+++ b/files/usr/share/elasticsearch/bin/elasticsearch.in.sh
@@ -1,0 +1,88 @@
+#!/bin/sh
+
+# check in case a user was using this mechanism
+if [ "x$ES_CLASSPATH" != "x" ]; then
+    cat >&2 << EOF
+Error: Don't modify the classpath with ES_CLASSPATH. Best is to add
+additional elements via the plugin mechanism, or if code must really be
+added to the main classpath, add jars to lib/ (unsupported).
+EOF
+    exit 1
+fi
+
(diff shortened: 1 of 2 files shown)


karmi:
Hi @rhoml, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/puppet-elasticsearch/pull/828.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

rhoml:
@karmi I've amended the commit

tylerjl:
Hi @rhoml,

It looks like the `elasticsearch.in.sh` file may have issues, as all of the acceptance tests break with the change - I'd suggest trying out running them locally with debugging output enabled (`BEAKER_debug=true`) to get additional output.

Beyond that though, I'd like to get a better idea of your motivating use case for this PR. The ability to change JVM flags is exposed through [the `jvm_options` parameter](https://github.com/elastic/puppet-elasticsearch/tree/f808dc4a085f4ae3c7fa604c8cfc7aae96062b95#java-installation) and is the recommended way to change those flags, and `init_defaults` exists if any other environment variables need to be changed. In general, I'd like to avoid overwriting the upstream package files when we can since it avoids a whole class of bugs that crop up with differing Elasticsearch versions. Are there some other examples that using the `jvm.options` parameter wouldn't solve?

rhoml:
@tylerjl disregard this PR, I completely overlooked at it as we are running this module's version from +6 months ago

PR #834, opened 2017-05-28
Reset potentially inherited defaults (since the mode is not always sp…

alexfouche:
Reset potentially inherited defaults (since the mode is not always specified in below items

For example "${elasticsearch::homedir}/lib" was stripped of 'other' permissions, so my Elasticsearch was unable to start

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -31,6 +31,13 @@ class elasticsearch::config {
     cwd  => '/',
   }
 
+  # Reset potentially inherited defaults (since the mode is not always specified in below items, for example "${elasticsearch::homedir}/lib" )
+  File {
+      owner       => undef,
+      group       => undef,
+      mode        => undef,
+  }
+
   if ( $elasticsearch::ensure == 'present' ) {
 
     file {


tylerjl:
Thanks @alexfouche. My concern here is that "fixing" resource defaults that occur above the elasticsearch module essentially means that the module is now attempting to proactively fix what the user is setting as resource defaults, and that is essentially unbounded. I'm definitely willing to consider the change, but it feels like this is less of a change the module needs to make and more of a tweak to the Puppet manifest that is making very broad resource defaults that impact other modules (for example, this module only uses resource defaults within the scope of resources it manages).

alexfouche:
Hi,
In fact, the fix is simply due to the fact that there is an item in this
file to manage "${elasticsearch::homedir}/lib", which specifies owner and
group to root, but does not specify a mode ​which is open to anyone,
especially the $elasticsearch::elasticsearch_group
Other ressources there may not specify the mode, but at least the user or
group is set for Elasticsearch, which can have after all keep access.

So an alternate fix could simply be sure that the mode is specified there,
but the recurse makes it not ideal

"${elasticsearch::homedir}/lib":
    ensure  => 'directory',
    group   => '0',
    owner   => 'root',
    recurse => true;

tylerjl:
While I agree with you that the `recurse` parameter doesn't make it ideal, I think the module has to accept that since it decides to manage the `lib` dir in the first place. Because the `file` resource can accept custom modes that aren't simply hex codes, what if we added the following to the `file` resource for the `lib` directory?

```puppet
recurse => true,
mode => 'o+Xr',
```

That should the directory is `r-x` and all the files beneath it are `r--`, thoughts?

fatmcgav:
Hi there,

Thanks again.

Issue #836, opened 2017-05-30
ES5 on CentOS 7 fails due to 'max number of threads is too low'


fatmcgav:
Elasticsearch 5.4 is failing to start on a freshly installed CentOS 7 VM with the following error:

>  max number of threads [1861] for user [elasticsearch] is too low, increase to at least [2048]

It appears that Systemd ignores the values specified in 'limits.conf', so need to add `LimitNPROC` to systemd unit file. 

PR incoming to implement.

tylerjl:
Implemented in #837, will go out with next release.

PR #837, opened 2017-05-30
Add support for setting NPROC value in systemd unit

fatmcgav:
As per #836, Elasticsearch 5 fails to start on CentOS 7 due to 'max number of threads is too low'. 

This PR adds support for over-riding the 'NPROC' value.

- [ ] Tests pass

### DIFF
--- a/files/etc/init.d/elasticsearch.systemd.erb
+++ b/files/etc/init.d/elasticsearch.systemd.erb
@@ -30,7 +30,7 @@ StandardError=inherit
 # Specifies the maximum file descriptor number that can be opened by this process
 LimitNOFILE=<%= @resource[:nofile] %>
 
-# Specifies the maximum number of threads that can be opened by this process
+# Specifies the maximum number of processes
 LimitNPROC=<%= @resource[:nproc] %>
 
 # Specifies the maximum number of bytes of memory that may be locked into RAM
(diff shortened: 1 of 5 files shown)


tylerjl:
Thanks @fatmcgav - just to give you some history, this module tries to keep in-sync with the upstream unit from the Elasticsearch packages to avoid any discrepancies between the two, so I raised this question to the Elasticsearch team to see whether this was something that should be added to the stock systemd unit (that we can then mirror here).

[There's now a PR to add this to the upstream Elasticsearch service unit](https://github.com/elastic/elasticsearch/pull/24970) so I think this is definitely a good change. I reproduced the one Travis CI failure locally; apparently version 6 of Elasticsearch requires a minimum limit of 4096 to avoid bootstrap errors, so I think setting the default on the puppet side to 4096 would be a good idea and get everything passing.

fatmcgav:
@tylerjl Cheers for the update. 

With regards to 2048/4096 difference, should this module handle the difference between 5.x and 6.x? Or is setting it to the 6.x default sufficient?

tylerjl:
@fatmcgav I think using the 6.x default is fine. I'm trying to limit version checking stuff in the module since it's so hairy and using 4096 should work fine with pre-6.x (and if it doesn't, CI will catch it).

fatmcgav:
@tylerjl OK, so I've updated the default to 4096, and also matched the wording of the systemd unit with the upstream changes... 

Now just to wait for TravisCI to run through...

tylerjl:
Thanks for the contribution @fatmcgav, really appreciate the work. This'll go out with the next release of the module.

PR #870, opened 2017-09-04
systemd: restart service on failure

uvNikita:
Restart=on-failure option will tell systemd to restart the service if it exited with non-zero exit code.

See [Restart](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=) option in systemd manual. In particular, it recommends the following:

> Setting this to on-failure is the recommended choice for long-running services, in order to increase reliability by attempting automatic recovery from errors.

The default value for [StartLimitIntervalSec](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#StartLimitIntervalSec=) will limit service to be restarted not more than 5 times within 10 seconds.

There is also a possibility to set this parameter optionally, but I think it would be a good default behavior.

- [x] Tests pass

### DIFF
--- a/files/etc/init.d/elasticsearch.systemd.erb
+++ b/files/etc/init.d/elasticsearch.systemd.erb
@@ -18,6 +18,8 @@ ExecStart=/usr/share/elasticsearch/bin/elasticsearch \
                                                 -p <%= @resource[:pid_dir] %>/elasticsearch-<%= @resource[:instance] %>.pid \
                                                 <%= opt_flags.join(' ') %>
 
+Restart=on-failure
+
 # StandardOutput is configured to redirect to journalctl since
 # some error messages may be logged in standard output before
 # elasticsearch logging system is initialized. Elasticsearch


tylerjl:
The systemd service template in this module is intentionally the same as the upstream one that the Elasticsearch service ships with, because changing the Puppet one can introduce inconsistencies with behavior in the default setup. See some related discussion [here on the elasticsearch repo](https://github.com/elastic/elasticsearch/issues/25425) and the [related discussion in the Ansible module](https://github.com/elastic/ansible-elasticsearch/issues/371).

With that in mind, I think the best option would probably be to drop a file as a service override, like the following (disclaimer: not tested, may or may not be right):

```puppet
file { "/etc/systemd/system/elasticsearch-${instance}.service/override.conf":
  ensure => file,
  content => "[Service]\nRestart=on-failure\n",
}
```

Thoughts?

uvNikita:
Thank you for pointing to the relevant discussions. Yes, it seems like the same workaround is currently advised for Chef and Ansible, so we should probably do the same in Puppet. Considering this, closing the PR.

Issue #877, opened 2017-09-26
Plugin Install Fails on 6.0.0-beta2


dfwarden:
* Module version: master
* Puppet version: 4.10.8
* OS and version: Oracle Linux 7.4 (Centos 7 and Ubuntu 1604 also failed)

## Bug description

The `elasticsearch::plugin` class fails against Elasticsearch 6.0.0-beta2 with the following error:

```
Error: Could not set 'present' on ensure: Failed to install plugin. Received error:
#<Puppet::ExecutionFailure: Execution of '/usr/share/elasticsearch/bin/elasticsearch-plugin install --batch x-pack'
returned 1: ES_PATH_CONF must be set to the configuration path> at/etc/puppetlabs/code/modules/elasticsearch/manifests/plugin.pp:129
```

I first noticed this issue on a dev VM using module version 5.4.3 and was able to recreate it in the snapshot acceptance tests on master with [a quick and dirty addition](https://github.com/dfwarden/puppet-elasticsearch/blob/9c4e1c70c773b1f6a182bf08394810a065f4e0b7/spec/acceptance/snapshot.rb#L82-L112).

I poked around the history a bit and noticed that the check for `ES_PATH_CONF` is in the new `elasticsearch-env` script, which `elasticsearch-plugin` calls first. It looks like `ES_PATH_CONF` started being removed from the per-os "defaults" file by [this commit](https://github.com/elastic/puppet-elasticsearch/commit/6f4c7dd1b2db73d4721630491b9ae10517e303bb) in favor of being managed at the instance level. To my naive eye one simple solution would be to leave `ES_PATH_CONF` in the "defaults" file, but I'm sure there are reasons for its removal that I'm not aware of. Another fix might involve adding `ES_PATH_CONF` support to the [with_environment function](https://github.com/elastic/puppet-elasticsearch/blob/01732062157120a197dce0135dd4f65219115b79/lib/puppet/provider/elastic_plugin.rb#L209-L234). Or maybe I'm misunderstanding something, in which case I apologize.

tylerjl:
Thanks @dfwarden, from a glance I think your assessment is correct. `ES_PATH_CONF` is removed from the defaults file because `elasticsearch-env` overrides any local environment variables, so it needs to be removed to let instances run in their own conf directories.

I think all that needs to be done is do adapt the plugin provider a bit to set the environment variable appropriately; this is already done with other variables like `ES_JAVA_OPTS` as you mention.

PR #898, opened 2017-11-28
export ES_PATH_CONF env variable in Red Hat init script

lukebigum:
This minor patch exports the ES_PATH_CONF variable inside the Red Hat init script to solve this problem:

`[root@localhost ~]# /etc/init.d/elasticsearch-es-01  start
Starting elasticsearch-es-01: ES_PATH_CONF must be set to the configuration path
                                                           [FAILED]
`

Somewhat similar to #877 but related to the starting of services, not plugins.

- [x] Tests pass

### DIFF
--- a/files/etc/init.d/elasticsearch.RedHat.erb
+++ b/files/etc/init.d/elasticsearch.RedHat.erb
@@ -53,6 +53,7 @@ export ES_JAVA_OPTS
 export ES_JVM_OPTIONS
 export JAVA_HOME
 export ES_INCLUDE
+export ES_PATH_CONF
 
 lockfile=/var/lock/subsys/$prog
 


lukebigum:
Considering it's part of elasticsearch that's complaining, I'd guess "probably", but haven't checked. Odd that an acceptance test hasn't caught that the service is not actually starting.

tylerjl:
Indeed, I'm going to look and see if I can refactor some of the acceptance tests to get failures to catch this reliably.

tylerjl:
Thanks for highlighting the bug, @lukebigum. I branched off of your branch and 1) fleshed out the snapshot tests so we're _actually_ covering additional Debian-based distributions, 2) got failures for those systems that run SysV init, and 3) replicated your fix into the Debian SysV init templates. Any future failures that fall into that bucket (Debian-based SysV init scripts for latest versions of Elasticsearch) should now reliably fail in CI.

Thanks for catching and fixing this one!

PR #902, opened 2017-12-01
option to set JAVA_HOME for elasticsearch plugins

lukebigum:
This is an adaption of PR #754 that's recently rebased, and basically piggy backs off the existing Environment modification code you've got in the plugin provider to also allow overriding JAVA_HOME, for those non-standard Java installs.

- [x] Tests pass


lukebigum:
I agree, Absolutepath.  I've added some unit tests as well, including a test to check we don't override an existing JAVA_HOME environment if it's already set and the param is not set (couldn't figure out how to get a nil value to work, so am testing empty string instead).

tylerjl:
Thanks @lukebigum, looks great!

PR #911, opened 2017-12-21
Add elasticsearch_snapshot_repository resource.

### LINKED ISSUE #896: Add Snapshot Repository support
Is there any appetite for adding support for managing '[Snapshot Repository](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-snapshots.html)' configuration to this module?

I've added an '[elasticsearch_snapshot_repository](https://github.com/weareact/puppet-elasticsearch/commit/ace1be16ba0dce275f7bdca9c9d109aa880a0cd1)' resource to our fork, albeit on the 0.x branch as that's what we're currently using. 
However I could quite easily re-work against master if it was worth it?

Thoughts?

fatmcgav:
Closes #896  

TODO: 
- [x] Implement functionality
- [x] Implement tests
- [x] Add documentation

- [x] Tests pass


fatmcgav:
@tylerjl So I've made a start on adding the new resource... 

The thing I'm struggling to get my head around atm is how to handle the data manipulation.
Looking at other providers, I presume I'm going to need to add a lamba or two to format the data, like the [elasticsearch_index](https://github.com/elastic/puppet-elasticsearch/blob/master/lib/puppet/provider/elasticsearch_index/ruby.rb#L12) provider.

fatmcgav:
OK, the more I look at this, the less confident I am that this resource is going to fit into the 'standard' REST Resource format... 

It looks like the `elasticsearch_template` and `elasticsearch_index` resources have a single property, which allows this code to work: https://github.com/elastic/puppet-elasticsearch/blob/master/lib/puppet/provider/elastic_rest.rb#L104-L110

So it looks like either:
- the `snapshot_repository` type needs to model a single `settings` property, though not sure how that would handle the `type` param.
- the `api_objects` method needs to be updated to handle resources with multiple properties/params... 

Thoughts?

tylerjl:
Hey @fatmcgav, thanks for contributing 👍 🚀 

I looked at the snapshot API and existing providers and agree with your assessment.

I'm not married to the way I wrote the REST code, so if there's changes necessary to support multiple parameters, then let's go for it. I'm sure we'll eventually support other REST-based API resources with > 1 parameter down the road, so I think it makes sense to do it properly.

Looking at the implementation in [`elastic_rest.rb`](https://github.com/elastic/puppet-elasticsearch/blob/6b4a6658fc942c4b0fc113ced1e9355a8b1da912/lib/puppet/provider/elastic_rest.rb), maybe the cleanest way forward would be to have a mechanism to forego an explicit "metadata" parameter and just let the provider merge in the hash in api_objects. That is, rather than constructing the hash that we eventually pass into `new` with `metadata => process_metadata ...`, maybe we could set the `name`, `provider`, and `ensure` values in a hash and `merge` in the `api_object` in that [`map` call](https://github.com/elastic/puppet-elasticsearch/blob/6b4a6658fc942c4b0fc113ced1e9355a8b1da912/lib/puppet/provider/elastic_rest.rb#L104). Then the resource will have a `settings` and `type` parameter without needing to explicitly state that they exist in the provider code, probably just for the native type. Thoughts?

fatmcgav:
@tylerjl I think I can see what you're getting at...

Let me see what I can work up... 😀

fatmcgav:
@tylerjl I've just pushed a couple of additional commits, one which adds `process_body` and `generate_body` functions to `elasticsearch_rest.rb`, which allows providers to over-ride the default behaviour if the resource is more complex. 

I've also updated the `elasticsearch_snapshot_repository` provider and supporting tests. However there's still one failing test that I need to debug.
```
Failures:

  1) Puppet::Type::Elasticsearch_snapshot_repository::ProviderRuby flush creates snapshots
     Failure/Error: let(:resource) { Puppet::Type::Elasticsearch_snapshot_repository.new props }

     NoMethodError:
       undefined method `unmanaged' for Puppet::Type::Elasticsearch_snapshot_repository::Compress:Class
     Shared Example Group: "REST API" called from ./spec/unit/provider/elasticsearch_snapshot_repository/ruby_spec.rb:71
     # ./.vendor/ruby/2.3.0/gems/puppet-4.10.9/lib/puppet/type.rb:930:in `block in managed?'
     # ./.vendor/ruby/2.3.0/gems/puppet-4.10.9/lib/puppet/type.rb:928:in `each'
```

tylerjl:
Nice, the approach makes sense to me, I like it. The error message is certainly weird though, I haven't run into that one before 🤔

fatmcgav:
OK, I think we're getting there now :) 

I've added a `snapshot_repository` defined type which mirrors the `template` version, added some unit and acceptance tests on the resource, confirmed it works in anger. 

Just need to add some docs to the README, and tidy up some of the commits I think :) 

@tylerjl Thoughts?

lib/puppet/type/elasticsearch_snapshot_repository.rb
```diff
@@ -0,0 +1,52 @@
...
+  newproperty(:max_restore_rate) do
+    desc 'Maximum Restore rate'
+  end
+
+  newproperty(:max_snapshot_rate) do
+    desc 'Maximum Snapshot rate'
+  end
+
+  # rubocop:disable Style/SignalException
+  validate do
```
tylerjl: Was this leftover?

manifests/snapshot_repository.pp
```diff
@@ -0,0 +1,104 @@
...
+#   Protocol that should be used to connect to the Elasticsearch API.
+#
+# @param api_timeout
+#   Timeout period (in seconds) for the Elasticsearch API.
+#
+# @param repository_type
+#   Snapshot repository type.
+#
+# @param compress
+#   Location of snapshots. Mandatory
```
tylerjl: Accidentally swapped this with location?

fatmcgav: Good spot, corrected.

lib/puppet/type/elasticsearch_snapshot_repository.rb
fatmcgav: Yeh, currently there's no validation... 

I'm wondering if there should be some though... Thoughts?

manifests/snapshot_repository.pp
```diff
@@ -0,0 +1,104 @@
...
+  Optional[Stdlib::Absolutepath]  $api_ca_path             = $elasticsearch::api_ca_path,
+  String                          $api_host                = $elasticsearch::api_host,
+  Integer[0, 65535]               $api_port                = $elasticsearch::api_port,
+  Enum['http', 'https']           $api_protocol            = $elasticsearch::api_protocol,
+  Integer                         $api_timeout             = $elasticsearch::api_timeout,
+  Boolean                         $compress                = true,
+  Optional[String]                $chunk_size              = undef,
+  Optional[String]                $max_restore_rate        = undef,
+  Optional[String]                $max_snapshot_rate       = undef,
+  Optional[String]                $repository_type         = undef,
```
fatmcgav: I guess this could technically be an `Enum`, as there's only a limited number of supported `repository types` in Elasticsearch. Thoughts?

lib/puppet/type/elasticsearch_snapshot_repository.rb
tylerjl: Maybe it could verify that the snapshot repository has a `location` set, since that's always required AFAIK.

manifests/snapshot_repository.pp
tylerjl: I don't feel too strongly about it; there is a fairly small set of repository plugins but I'd be fine leaving it as-is since users may be using any sort of repository plugin, including custom ones they've potentially written that we'd be unable to predict anyway.

fatmcgav:
@tylerjl OK, I've added validation logic to the type, and squashed the mirriad of commits down into 5 logical commits :) 

Do I need to update the CHANGELOG, or will you do that after merging?

Otherwise, this is good to go :) 
Phew!

tylerjl:
@fatmcgav feel free to put the cherry on top with a CHANGELOG message of your choice. I'll make some time tomorrow to merge and cut a new release since this feels like a worthwhile feature to release for.

fatmcgav:
@tylerjl The cherry is on the top! :D 🎉

fatmcgav:
Wooh!!!

PR #925, opened 2018-04-05
Stop default service starting

### LINKED ISSUE #924: Default service that is masked is also started
* Module version: 6.2.2
* Puppet version: 4.10.9
* OS and version: Ubuntu 16.04

## Bug description

The default service that should be masked is also started. When I run Puppet I end up with any instances defined with `elasticsearch::instance` being started, along with the default `Service[elasticsearch]` which comes with the package and should not be started.

From the Puppet run:
```
Notice: /Stage[main]/Elasticsearch::Config/Service[elasticsearch]/ensure: ensure changed 'stopped' to 'running'
```

I believe that https://github.com/elastic/puppet-elasticsearch/blob/master/manifests/config.pp#L100 should include something like:
```
enabled => false
```

jarro2783:
Fixes #924. Stops the default service that is masked from being started by Puppet.

- [ ] Rebased/up-to-date with base branch
- [ ] Tests pass

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -69,8 +69,6 @@ class elasticsearch::config {
         ensure => 'absent';
       '/etc/elasticsearch/log4j2.properties':
         ensure => 'absent';
-      '/etc/init.d/elasticsearch':
-        ensure => 'absent';
     }
 
     if $elasticsearch::pid_dir {
@@ -98,7 +96,13 @@ class elasticsearch::config {
     if ($elasticsearch::service_provider == 'systemd') {
       # Mask default unit (from package)
       service { 'elasticsearch' :
+        ensure => false,
         enable => 'mask',
+      }
+    } else {
+      service { 'elasticsearch':
+        ensure => false,
+        enable => false,
       }
     }
 
--- a/spec/acceptance/002_class_spec.rb
+++ b/spec/acceptance/002_class_spec.rb
@@ -29,6 +29,11 @@ describe '::elasticsearch' do
       end
     end
 
+    describe service('elasticsearch') do
+      it { should_not be_running }
+      it { should_not be_enabled }
+    end
+
     describe service(test_settings['service_name_a']) do
       it { should be_enabled }
       it { should be_running }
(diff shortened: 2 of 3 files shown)


jarro2783:
I'm not sure why it fails the test for the service should not be running on some systems and not others.

tylerjl:
This likely because those systems that pass tests are systemd-based distributions. In order for the other distributions to enforce the same disabled services, the SysV-init style services would need to be disabled/stopped as well.

jarro2783:
Should I drop that check for the other distributions then? Or would it be useful to put `ensure => false` in for `init.d` systems too?

tylerjl:
Keeping the rest of the init systems consistent would be important, so ensuring their absence would be a good thing to do. That's one aspect of the change here that you could add, I think it would include removing the `ensure => absent` that the module currently does to the init file and then removing the conditional that only stops the service for systemd-based systems. Note that `mask`ing a service will also need to be handled for distros that aren't on systemd.

There's a broader question about how this may affect users that upgrade, however: if they upgrade from the current version of the module to a release that incorporates these changes, I _assume_ that SysV-based systems will fail to disable the stock service since the file will have been removed from the `absent` setting on the file and may fail in that step. Not sure how to best handle that...

jarro2783:
Is that going to be an actual problem. Presumably anyone on a SysV system will not still have a running Elasticsearch service, since it would have failed due to a lack of `/etc/elasticsearch/elasticsearch.yml`. So when they do an upgrade Puppet won't actually need to stop anything.

tylerjl:
I ran a test going from a setup on the master branch to this PR and didn't see any problems, so it looks like it shouldn't behave weirdly for users who upgrade. I think to get the tests happy you may just need to remove the line that sets `ensure => absent` on the init.d service file so `service` commands still work, which is a good idea anyway so that the module can correctly stop/disable the service instead of just failing to see the service at all.

jarro2783:
Now it's failing with a seemingly unrelated error. The `intake` tests run fine on my machine. Any ideas about what's going on?

tylerjl:
Looks like there's [an upstream fix for puppetlabs_spec_helper](https://github.com/puppetlabs/puppetlabs_spec_helper/pull/237) that's still unreleased - to help the tests pass, I committed a fix to `master` that pulls in the ref that implements the fix. If you merge/rebase onto master your branch should get the fix. Sorry about that, kind of a pain but I think we can workaround the breaking changes that happened upstream.

jarro2783:
I think the `opensuse` test has been failing for a while on master, otherwise it all passes now.

Issue #940, opened 2018-06-01
Is there a way to create a document in ES with this module?


nick-george:
## Feature Description
Hi, 

This module already has very rich functionality, and for that I'm very grateful! Have you considered adding the ability to create a document in ES using this module? Does this feature already exist? I've had a look at the custom types, and can't see anything that would do this. 

It would be awesome if we had a resource similar to the puppet "file" resource so we could drop documents in. I'm not super familiar with custom types and providers, but to implement this, could a new provider be written for the "file" resource? 

Thanks for your time!
Nick George

tylerjl:
Some of the types and providers in this module like index settings, templates, and pipelines directly manipulate the Elasticsearch API to manage documents that can be represented as Puppet hashes. Whether this was a good idea at was somewhat as a gray area, as letting config management bleed over into manipulating a data store is a pretty unconventional use of Puppet. I think it's been generally useful though; as empowering the module to control all the settings that users may need and not just configuration files makes the module more complete from an end-user perspective.

I'm not sure about explicitly managing documents, though. I've sort of referenced the [Puppetlabs MySQL module](https://forge.puppet.com/puppetlabs/mysql/reference) as a bit of a guiding example of _what_ sorts of resources this module should handle, and though there are certainly parallels between the two right now (for example, the mysql module can manage a database schema in the same way this module can handle index settings), managing records/actual data is something kind of unusual and not something I've seen in it or any other modules that deal with data stores (mongo, postgres, etc.). 

Perhaps some examples of a) what these resource would look like and b) what the use case is (i.e. what types of documents and what the intent of managing them via Puppet is) could help to clarify whether it'd be a good fit.

nick-george:
Hi Tyler, 

I'm probably teaching you to suck eggs here, but one of the cool features of ES is that it can be used to store things like configuration data. For example, Kibana's saved objects like dashboards, visualisations and searches are all stored in Elasticsearch. I'm also considering moving some configuration (for a custom Kibana plugin) that I currently store on disk into Elasticsearch. It would be awesome if I could just write this config data straight into ES using this module. 

There are a few small advantages to being able to store "files" in ES using the Puppet module. 
1. It becomes possible configure software (like Kibana) completely, using Puppet. I understand that the Kibana module might one day have the ability to set dashboards etc via its module, and will probably do so using the Kibana API. Who knows, maybe they'd leverage this feature to implement the restoration of dashboards etc. 
[...]

PR #972, opened 2018-08-06
Elasticsearch document defined type

nick-george:
- [Y] Signed the [CLA](https://www.elastic.co/contributor-agreement/) (if not already signed)
- [Y] Rebased/up-to-date with base branch
- [N] Tests pass
- [N] Updated CHANGELOG.md with patch notes (if necessary)
- [Y] Updated CONTRIBUTORS (if attribution is requested)

Hi @tylerjl and co,

I have created an elasticsearch_document defined type (and associated provider). This implements the defined type I was requesting in #940. I have not (yet) updated the docs, but it works much like the other native types in this module. This is my first ever attempt at a Puppet type & provider, so very happy to take feedback on how it could be improved. 

Here are a few notes: 
- I did not write an "instances" method for this type, I get the impression that this needs to return all documents that could be considered resources by Puppet. I figure the output from this could be unbounded and would likely kill Puppet. 
- My code does not use the deep_to_s method, and leaves everything as specified (e.g booleans stay as booleans, numbers stay as numbers). 
- I wrote some tests for the "type", but not for the "provider". It seems that the helper script (spec/helpers/unit/provider/elasticsearch_rest_shared_examples.rb) uses the instances method, which I have not implemented. 
- I couldn't get one of the "REST API types" tests working (PSON one) as it seems to expect my code to call deep_to_s, which my code does not do. I don't know how to selectively turn off one test. 
- The insync? method is calling deep_implode. Not sure if this is the right thing to do for documents. Could I just do a direct hash comparison e.g is == should? 
- I have not updated the README.md 

It seems to be working well, and is allowing me to define documents in Puppet (and hiera) and have Puppet insert them into ES.


tylerjl:
Thanks for the contribution @nick-george! I'll take a look soon.  Also /cc @fatmcgav as he's also helping out with maintenance of the module.

Jenkins, test this please.

Issue #996, opened 2018-11-05
transport port fact


kobybr:
Would be useful to have the transport.tcp.port set as a fact along with the current http.port values.

PR #999, opened 2018-11-11
Add transport port fact

kobybr:
Resolves issue #996


tylerjl:
Jenkins, test this please

kobybr:
Any word on this being merged?

lib/facter/es_facts.rb
```diff
@@ -110,10 +116,29 @@ def self.run
             end
           end
           add_fact(key_prefix, 'plugins', plugin_names.join(','))
+
+          nodes_data['http']['bound_address'].each { |i| http_bound_addresses << i }
+          nodes_data['transport']['bound_address'].each { |i| transport_bound_addresses << i }
+          transport_publish_addresses << nodes_data['transport']['publish_address']
+          transportports << nodes_data['settings']['transport']['tcp']['port']
```
tylerjl: Looking at an example response for a node's settings from the node stats endpoint, I see the following under `.settings.transport`:

```
{
  "type": {
    "default": "netty4"
  }
}
```

Am I correct in thinking that without a `tcp` key, accessing that `nil` value with `['port']` will throw an error when trying to access a non-existent key on `nil`?

kobybr: Good catch!   I've pushed new code with what I believe will handle that.

kobybr: FYI.   This original pull request was intended add additional facts to support those of us who manage firewall/iptables via puppet.

tylerjl:
Thanks @kobybr - the one failed test is unrelated and the nil check looks safer, so this looks good to merge! :+1:

kobybr:
Thanks!

PR #1003, opened 2018-11-27
manifests/plugin: reorder plugin notify

GiedriusS:
In some rare cases Puppet starts to think that we are redefining the
elasticsearch_plugin entity when it is used before the actual resource
declaration and then applying the catalog fails:

Error: Could not retrieve catalog from remote server: Error 500 on SERVER:
Server Error: Evaluation Error: Error while evaluating a Resource Statement,
Evaluation Error: Error while evaluating a Resource Override, Could not
autoload puppet/type/elasticsearch_plugin: Attempt to redefine entity
'http://puppet.com/2016.1/runtime/type/elasticsearch_plugin'. Set at unknown
location at
/etc/puppetlabs/code/environments/main/modules/elasticsearch/manifests/plugin.pp:94:5
at
/etc/puppetlabs/code/environments/main/core/profiles/manifests/foo/elasticsearch.pp:104
on node
foo.bar.540455029edf40888d96132b81ff8db7f1d553bc

This output is from the older plugin version but it happens with the newer one
as well since the same code block exists in both versions.Reordering and putting
the notify block after declaring the resource fixes this problem.

- [x] Tests pass
- [x] Updated CONTRIBUTORS


tylerjl:
Thanks for the fix @GiedriusS - it looks like the CI failures are unrelated, so this should be good to go. Thank you for contributing!

Issue #1002, opened 2018-11-22
Error: Could not find a suitable provider for elasticsearch_role/elasticsearch_user_roles


joernott:
* Module version: 6.3.3
* Puppet version: 4.10.8
* OS and version: Red Hat Enterprise Linux Server release 7.5 (Maipo)

## Bug description

When installing elasticsearch on an all-in-one development box together with many applications (and thus many modules), the following error messages pop up regularly and lead to a failed puppet run:

* Error: Could not find a suitable provider for elasticsearch_role
* Error: Could not find a suitable provider for elasticsearch_user_roles

This is due to the way, how these users are created, which triggers a bug in puppet when collecting resources.

We worked around this issue by using file resources for "${elasticsearch_home}/config/${instance}/x-pack/users", "${elasticsearch_home}/config/${instance}/x-pack/users_roles" and "${elasticsearch_home}/config/${instance}/x-pack/roles.yml" instead of using the defined types elasticsearch::role and elasticsearch::user for the file based users and defined our own defined types which create roles and users via curl calls to the API.

Attached examples are stripped down/simplified.

joernott:
Despite not using elasticsearch::user and elasticsearch::role any more, Puppet still shows an error when building a machine from scratch:
`27/11/2018 09:30:22     default: Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Virtual Query, Could not autoload puppet/type/elasticsearch_user_file: Could not autoload puppet/provider/elasticsearch_user_file/oss_xpack: no such file to load -- puppet/provider/elastic_parsedfile at /etc/puppetlabs/code/environments/vbox/modules/elasticsearch/manifests/init.pp:555:6 on node box-krn0-vbx-v01.sf-bk.de
`

shazi7804:
Could not autoload  +1

nod0n:
I had a similar Issue:
`Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Virtual Query, Could not autoload puppet/type/elasticsearch_user: Could not autoload puppet/provider/elasticsearch_user/esusers: no such file to load -- puppet/provider/elastic_user_command at /etc/puppetlabs/code/environments/production/modules/elasticsearch/manifests/init.pp:544:3 on node mynode`
[...]

Issue #982, opened 2018-09-11
Could not autoload puppet/provider/elasticsearch_user_file/oss_xpack


Celedhrim:
* Module version: 6.3.3
* Puppet version: 4.10.12
* OS and version:Debian GNU/Linux 9.5 (stretch)

## Bug description

Hello,

I have install elasticsearch 6.3.2 with puppet module version 6.2.2 but I have the scripts symlink bug.

So I have update elasticsearch module to 6.3.3.
I have made change on my manifest to take care of change ( oss + repository)

But now when I try to apply manifest : 
```
 ~# puppet agent -t
Info: Using configured environment 'my_build_branch'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Virtual Query, Could not autoload puppet/type/elasticsearch_user_file: Could not autoload puppet/provider/elasticsearch_user_file/oss_xpack: undefined method `oss_xpack_config' for Puppet::Type::Elasticsearch_user_file::ProviderOss_xpack:Class at /etc/puppetlabs/code/environments/my_build_branch/modules/elasticsearch/manifests/init.pp:555:6 on node my.server
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
```

Have you an idea ? this was introduce in 6.3.0.

tylerjl:
Hi @Celedhrim, there's a possibility that this came up when you updated the module on the puppetmaster and the native types and providers changed. Please try restarting the puppetmaster (and potentially any effected puppet agents) and try re-running puppet again.

aadiwarden:
We are seeing similar issues or our side.

Could not autoload puppet/type/elasticsearch_user_file: Could not autoload puppet/provider/elasticsearch_user_file/oss_xpack: undefined method `oss_xpack_config' for Puppet::Type::Elasticsearch_user_file::ProviderOss_xpack:Class at /etc/puppetlabs/code/environments/elastic6x_cfg/modules/elasticsearch/manifests/init.pp:545:6

Celedhrim:
> Hi @Celedhrim, there's a possibility that this came up when you updated the module on the puppetmaster and the native types and providers changed. Please try restarting the puppetmaster (and potentially any effected puppet agents) and try re-running puppet again.

I've tried it before opening the issue , both puppet server and puppet agent, but same error

[...]

PR #1020, opened 2019-02-26
Solution for error: Could not autoload puppet/type/elasticsearch_user

lkosz:
Hello

PR contains solution for error ```Could not find a suitable provider for elasticsearch_role/elasticsearch_user_roles
Could not autoload puppet/type/elasticsearch_user
Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Virtual Query, Could not autoload puppet/type/elasticsearch_user: Could not autoload puppet/provider/elasticsearch_user/elasticsearch_users: no such file to load -- puppet/provider/elastic_user_command```

### DIFF
--- a/lib/puppet/provider/elasticsearch_user/elasticsearch_users.rb
+++ b/lib/puppet/provider/elasticsearch_user/elasticsearch_users.rb
@@ -1,3 +1,5 @@
+$LOAD_PATH.unshift(File.join(File.dirname(__FILE__), '..', '..', '..'))
+
 require 'puppet/provider/elastic_user_command'
 
 Puppet::Type.type(:elasticsearch_user).provide(


tylerjl:
Hi @lkosz - the problem you mention this as a solution for talks about using `require_relative` and refers to a different provider (`elasticsearch_user_roles`) instead of `elasticsearch_user` as you've changed here. Is `$LOAD_PATH` changes the right change, and if so, does it need to be made to the rest of the files that include relative files in `lib/`?

lkosz:
@tylerjl hmmm yes, you're right, I've pasted wrong error :) This PR solve following exact error on my side:
```Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Virtual Query, Could not autoload puppet/type/elasticsearch_user: Could not autoload puppet/provider/elasticsearch_user/elasticsearch_users: no such file to load -- puppet/provider/elastic_user_command```. I've changed PR title and description.
I'm preety sure I saw issue for it, but can't find it now. 
It's related to both of issues: https://github.com/elastic/puppet-elasticsearch/issues/982 and https://github.com/elastic/puppet-elasticsearch/issues/1002 and I think attachment of such line to other libraries may be helpful :)

tylerjl:
@lkosz I agree and think that preemptively adding that line to other library files would be good here, otherwise I think that other users (and yourself) will probably just run into additional exceptions later in the run when Puppet tries to load files beyond the one changed here. Could you add that fix to the other libraries files to ensure that it's fixed for everything?

daisukixci:
Are there news about this PR ?

smasa90:
https://github.com/elastic/puppet-elasticsearch/issues/982#issuecomment-503791056

tylerjl:
Closing in favor of #1036 which covers more files :+1:

PR #1024, opened 2019-03-18
[ci] add JJBB definitions

jonahbull:
Add JJBB definitions to the repo so they can be managed here.

* Enable auto-generation of `inject-passwords` section

* Move `VAULT_PATH` to a plain environmental variable since we don't
currently support auto-generation of `inject-passwords` and "extra"
attributes.

* Note that the publisher `infra-root+build@elastic.co` is now the
default, so it'll be used for any jobs that don't override it. Let me
know if that's not what we want and I'll update.

- [ ] Tests pass

### DIFF
--- a/.ci/jobs/defaults.yml
+++ b/.ci/jobs/defaults.yml
@@ -1,0 +1,53 @@
+---
+
+##### GLOBAL METADATA
+
+- meta:
+    cluster: devops-ci
+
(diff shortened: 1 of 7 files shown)


tylerjl:
Thanks! The changes look right; unfortunately the tests are in a state of disrepair at the moment but those failures are unrelated to the yaml here.

Issue #1032, opened 2019-04-22
Default jvm.options incompatible with bundled JDKin elasticsearch 7


6uhrmittag:
* Module version: mod 'elastic-elasticsearch', '6.3.3'
* Puppet version: PuppetBolt 1.10.0 
* OS and version: Ubuntu 16.04.6 LTS

## Bug description

The Module defaults in `jvm.options` are not compatible with the bundled Java in elasticsearch 7. Elasticsearch won't start with the example Puppet manifest.

### Reproduce

With:
````
elasticsearch::instance { 'es-01':
  config      => {
    #  'xpack.monitoring.collection.enabled' => true,
    'network.host' => '127.0.0.1',
    'http.port'    => '8200',
  }
}
````
Result: Elasticsearch won't start.

Syslog:
````
Apr 22 18:18:49 ip-172-31-35-88 elasticsearch[18037]: Unrecognized VM option 'PrintGCApplicationStoppedTime'
Apr 22 18:18:49 ip-172-31-35-88 elasticsearch[18037]: Error: Could not create the Java Virtual Machine.
Apr 22 18:18:49 ip-172-31-35-88 elasticsearch[18037]: Error: A fatal exception has occurred. Program will exit.
Apr 22 18:18:49 ip-172-31-35-88 systemd[1]: elasticsearch-es-01.service: Main process exited, code=exited, status=1/FAI
LURE
Apr 22 18:18:49 ip-172-31-35-88 systemd[1]: elasticsearch-es-01.service: Unit entered failed state.
Apr 22 18:18:49 ip-172-31-35-88 systemd[1]: elasticsearch-es-01.service: Failed with result 'exit-code'.
````

The generated jvm.options:
````
root@xxxx:/etc/elasticsearch/es-01# cat jvm.options
# This file is managed by Puppet -- es-01
#
# Set the 'jvm_options' parameter on the elasticsearch class to change this file.

-Dfile.encoding=UTF-8
-Dio.netty.noKeySetOptimization=true
-Dio.netty.noUnsafe=true
-Dio.netty.recycler.maxCapacityPerThread=0
-Djava.awt.headless=true
-Djna.nosys=true
-Dlog4j.shutdownHookEnabled=false
-Dlog4j2.disable.jmx=true
-XX:+AlwaysPreTouch
-XX:+HeapDumpOnOutOfMemoryError
-XX:+PrintGCApplicationStoppedTime
-XX:+PrintGCDateStamps
[...]

PR #1021, opened 2019-02-27
Add node name to log4j config

jwils:
Elasticsearch automatically adds the node name but that will be removed in 7.0. This adds the node name explicitly to
the log4j config so it does not get lost when ES 7.0 is release.
Startup warning:
> Some logging configurations have %marker but don't have %node_name. We will automatically 
> add %node_name to the pattern to ease the migration for users who customize log4j2.properties 
> but will stop this behavior in 7.0. You should manually replace `%node_name` with 
> `[%node_name]%marker ` in these locations: ....

Pull request acceptance prerequisites:

- [ ] Signed the [CLA](https://www.elastic.co/contributor-agreement/) (if not already signed)
- [ ] Rebased/up-to-date with base branch
- [ ] Tests pass
- [ ] Updated CHANGELOG.md with patch notes (if necessary)
- [ ] Updated CONTRIBUTORS (if attribution is requested)

PR #1040, opened 2019-06-27
fix(compat): bare min. to get srv running on 7.x

smasa90:
Fixed jvm.options.erb to add required jvm options for 7.x
Fixed jvm.options.erb to change jvm options required only for openjdk 8
Added ES_PATH_DIR instance environment var
Fixed log4j2.properties.erb to fix Deprecation WARNINGS
Keeped Backward Compatibility

- [ x ] Signed the [CLA](https://www.elastic.co/contributor-agreement/) (if not already signed)
- [ x ] Rebased/up-to-date with base branch
- [ ] Tests pass


smasa90:
Seems to have a little dups here https://github.com/elastic/puppet-elasticsearch/pull/1021
Didn't see the PR before pushing this one.

manifests/instance.pp
```diff
@@ -162,6 +162,7 @@
   Boolean                            $ssl                           = false,
   Elasticsearch::Status              $status                        = $elasticsearch::status,
   Optional[String]                   $system_key                    = $elasticsearch::system_key,
+  Integer                            $repo_version                  = $elastic_stack::repo::version,
```
ernstae: This line will generate a warning if users aren't leveraging the `elastic_stack` puppet module, or explicitly passing in a value. 

```
Warning: Unknown variable: 'elastic_stack::repo::version'. 
```

Not sure if this is acceptable, or whether to consider it as taking an implicit dependency on the elastic_stack module

smasa90: You’re right, I’ve not thought as much on how to handle correctly this version condition. I’ll update this when I have the time ti dig into. Or if someone as suggestions, I’ll consider it. Marked this as WIP

smasa90:
You’re right, I’ve not thought as much on how to handle correctly this version condition. I’ll update this when I have the time ti dig into. Or if someone as suggestions, I’ll consider it. Marked this as WIP

smasa90:
Is that better like that ? Private var, no mandatory implicit dependency to elastic_stack module. Default to false if no version is provided nor in elastic_stack::repo::version nor in elasticsearch::version. Can default to true when community will be ready to default to 7.x with Puppet modules.

smasa90:
Can someone help me with that:

smasa90:
Ping @tylerjl ?

tylerjl:
Apologies for the delays here, I'm trying to get some more resources dedicated to some module updates and maintenance.

The short answer is that it gets _really_ challenging to make these sorts of version detection changes since the module is in total control of the jvm.options file - I'm hoping that dropping support for multiple instances will let us manage the jvm.options file on a line-by-line basis, so users will be able to add or remove lines as necessary (that match the format for the version of Elasticsearch they're running) and alleviate the need to populate the entire file at all.

smasa90:
How could we plan it ?
A solution can be to copy default jvm.options file on each instance and manage it line by line

sticky-note:
ping @tylerjl, I think we need this one way or an other.
If this module is not maintained any more, on which provisioner can we turn ourselves to get a maintained module ?

NITEMAN:
Rel #1032 
Any updates on this?

fatmcgav:
Hi there,

Thanks again.

PR #1074, opened 2020-03-10
Fix failing tests on master

fatmcgav:
This PR fixes up the failing tests on `master` branch, as seen here: https://travis-ci.org/elastic/puppet-elasticsearch/builds/660426140?utm_medium=notification&utm_source=github_status

As part of this fixup, removed tests for Elasticsearch `2.x`, and update to the latest `6.x` release.


Rakefile
```diff
@@ -137,7 +137,7 @@ beaker_node_sets.each do |node|
     "beaker:#{node}:acceptance", [:version, :filter] => [:spec_prep]
   ) do |task, args|
     ENV['BEAKER_set'] = node
-    args.with_defaults(:version => '6.2.3', :filter => nil)
+    args.with_defaults(:version => '6.8.6', :filter => nil)
```
jmlrt: any reason to not use 6.8.7?

fatmcgav: None other than `6.8.7` wasn't available when I did this initial commit... Will update in future round :)

PR #1075, opened 2020-03-11
Bump Beaker versions to support latest OS'

fatmcgav:
We need to bump the version of Beaker we're using to support the latest OS', such as EL8, Debian 10 and Ubuntu 18.04

This brought with it a few breaking changes that needed fixing.

Also had to bump the version of `puppetlabs-apt` and `puppetlabs-java` to support new OS'.


spec/spec_helper_acceptance.rb
```diff
@@ -1,18 +1,24 @@
...
 require_relative 'spec_utilities'
 require_relative '../lib/puppet_x/elastic/deep_to_i'
 require_relative '../lib/puppet_x/elastic/deep_to_s'
 
-def f
-  RSpec.configuration.fact
-end
+# def f
+#   RSpec.configuration.fact
+# end
```
jmlrt: can we remove this block if it isn't used?

spec/spec_helper_acceptance.rb
```diff
@@ -144,50 +150,18 @@ class { 'elasticsearch':
...
-  # We spawn a thread to print dots periodically while installing puppet to
-  # avoid inactivity timeouts in Travis. Don't judge me.
-  progress = Thread.new do
-    print 'Installing puppet..'
-    print '.' while sleep 5
-  end
+  # # Set the host to 'aio' in order to adopt the puppet-agent style of
+  # # installation, and configure paths/etc.
+  # host[:type] = 'aio'
+  # configure_defaults_on host, 'aio'
```
jmlrt: can we remove this block if it isn't used?

spec/spec_helper_acceptance.rb
```diff
@@ -274,10 +248,10 @@ class { "java" :
...
-    f['os']['release']['major'] == '6')) or \
-    f.dig 'os', 'distro', 'codename' == 'xenial'
-end
+# # Java 8 is only easy to manage on recent distros
+# def v5x_capable?
+#   (fact('os.family') == 'RedHat' and \
+#     not (fact('os.name') == 'OracleLinux' and \
+#     f['os']['release']['major'] == '6')) or \
+#     f.dig 'os', 'distro', 'codename' == 'xenial'
+# end
```
jmlrt: can we remove this block if it isn't used?

spec/spec_helper_acceptance.rb
fatmcgav: Yep, I'll clean these up in a future PR! :)

PR #1088, opened 2020-06-16
Add Elasticsearch 7 to test matrix

fatmcgav:
Add Elasticsearch `7.7.1` to the test matrix, and update supporting code and tests to support.

Also added Ubuntu `20.04` to the test matrix.


fatmcgav:
This one is ready for review now.

There is still one persistent failure on CentOS/Oracle Linux 7, which appears to be due to some weird issue with installing the RPM and file permissions. Related Elasticsearch issue: https://github.com/elastic/elasticsearch/issues/50631

Needs some more detailed debugging, but don't want that to hang this up too much.

jmlrt:
> Add Elasticsearch `7.7.1` to the test matrix, and update supporting code and tests to support.

Can this be updated to `7.8.0`?

> Also added Ubuntu `20.04` to the test matrix.

Ubuntu 20.04 is not yet on the [support matrix](https://www.elastic.co/support/matrix), I would prefer waiting to add it to puppet supported platforms.

fatmcgav:
I've switched the default 7.x version to `7.8.0` in 86b4fb7, and also removed Ubuntu 20.04.

Thanks for the feedback @jmlrt

spec/acceptance/nodesets/ubuntu-server-2004-x64.yml
```diff
@@ -0,0 +1,17 @@
+HOSTS:
```
jmlrt, line 1: can we remove also this file?

fatmcgav:
So I pushed the branch to the `elastic` repo in order to trigger some of the security tests, and they [all failed](https://travis-ci.org/github/elastic/puppet-elasticsearch/builds/704313414) :(

I'm going to debug further before merging.

fatmcgav:
OK, I think I've found the issue with the security tests (A different Vault secret path to the one I updated with `v7` license), and I'm running the job [again](https://travis-ci.org/github/elastic/puppet-elasticsearch/builds/704649551).

fatmcgav:
So the tests passed with the exception of the EL7 tests which I noted above.

Merging this one in and will work on the EL7 tests in a separate PR.

PR #1107, opened 2021-01-07
Fixed Service definition with custom service_name

StanvanHoorn:
When Elasticsearch is installed with a custom service_name, puppet will fail because the custom name is not used for the before relationship metaparameter.

```
class { 'elasticsearch':
  service_name        => 'elasticsearch_default',
}
```

results in:
`Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Could not find resource 'Service[elasticsearch]' in parameter 'before'`

- [X] Signed the [CLA](https://www.elastic.co/contributor-agreement/) (if not already signed)
- [X] Rebased/up-to-date with base branch
- [ ] Tests pass
- [X] Updated CHANGELOG.md with patch notes (if necessary)

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -69,7 +69,7 @@ class elasticsearch::config {
         owner  => 'root',
         group  => $elasticsearch::elasticsearch_group,
         mode   => '0660',
-        before => Service['elasticsearch'],
+        before => Service[$elasticsearch::service_name],
         notify => $elasticsearch::_notify_service,
       }
     } else {
@@ -77,7 +77,7 @@ class elasticsearch::config {
         incl    => "${elasticsearch::defaults_location}/elasticsearch",
         lens    => 'Shellvars.lns',
         changes => template("${module_name}/etc/sysconfig/defaults.erb"),
-        before  => Service['elasticsearch'],
+        before  => Service[$elasticsearch::service_name],
         notify  => $elasticsearch::_notify_service,
       }
     }


shamil:
I don't think this is the right approach, better have this in single place, I would change the `manifests/service.pp` where the service was defined and make it look like: 

```puppet
  service { 'elasticsearch':
    name   => $elasticsearch::service_name
    ensure => $_service_ensure,
    enable => $_service_enable,
  }
  ```

fatmcgav:
> When Elasticsearch is installed with a custom service_name, puppet will fail because the custom name is not used for the before relationship metaparameter.

I'd also be interested to understand the purpose behind changing the service name. With the recent removal of multi-instance support in the 7.x module release, we're trying to stick as closely as possible to the package defaults...

shamil:
> I'd also be interested to understand the purpose behind changing the service name. With the recent removal of multi-instance support in the 7.x module release, we're trying to stick as closely as possible to the package defaults...

Yes, I don't think we even able to change service name, since the module relies on the service unit file which was installed by the package manager, so we better remove the option to change the service

fatmcgav:
> > I'd also be interested to understand the purpose behind changing the service name. With the recent removal of multi-instance support in the 7.x module release, we're trying to stick as closely as possible to the package defaults...
> 
> Yes, I don't think we even able to change service name, since the module relies on the service unit file which was installed by the package manager, so we better remove the option to change the service name

This is where I was heading aswell...

@StanvanHoorn Let me know if that change would cause you any serious issues, or if changing the service name has just been a "because I can" moment...

StanvanHoorn:
The only reason was that we always named the service elasticsearch-default. So it has no impact for us if the service name cannot be changed anymore.

fatmcgav:
> The only reason was that we always named the service elasticsearch-default. So it has no impact for us if the service name cannot be changed anymore.

Thanks for confirming @StanvanHoorn 

I'll put a PR up later removing the `service_name` param and anything associated.

Issue #1113, opened 2021-03-03
When restart_config_change is set to false Elasticsearch is still restarted


scornelissen85:
* Module version: 7.0.0
* Puppet version: 6.x
* OS and version: RedHat 7

## Bug description

When the parameter `restart_config_change` on the class `elasticsearch` is set to `false`, Elasticsearch is still restarted by line 514: https://github.com/elastic/puppet-elasticsearch/blob/master/manifests/init.pp#L514 (there is an hardcoded notify from `Class['elasticsearch::config']` to `Class['elasticsearch::service']` which will result in a restart of the Elasticsearch service.

scornelissen85:
This seems to be  a duplicate of #1111, while that one resolved it by never notify Elasticsearch service, even when `restarton_config` is set to `true`.

oniGino:
i think its fixed in #1114 but i don't think a release has been made yet

smortex:
@oniGino 8.0.0 should include this. Can you try it and report the outcome? Thanks!

oniGino:
yes it is, thanks! i was still pulling from the elastic repo instead of voxpopuli

i think its safe to close this

smortex:
Awesome, thanks!

PR #1111, opened 2021-02-07
fix restart_on_change behavior

bovy89:
`restart_on_change` parameter is ignored. If config file is modified, the class `elasticsearch::config` always notify `elasticsearch::service` and the elasticsearch service is being restarted

PR #1114, opened 2021-03-03
Do not restart Elasticsearch on config change when restart_config_change is set to false

scornelissen85:
This change will fix that Elasticsearch service is not being restarted when `restart_config_change` is set to `false`. Without this change Elasticsearch was still restarted on config changes even when `restart_config_change` was set to `false`. See also issue #1113

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -511,7 +511,12 @@ class elasticsearch (
     # Installation, configuration and service
     Class['elasticsearch::package']
     -> Class['elasticsearch::config']
-    ~> Class['elasticsearch::service']
+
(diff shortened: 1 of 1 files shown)


crazymind1337:
@fatmcgav can we have this PR merged and released soon? I dont like to fork the module again for such a tiny issue.

shamil:
I think https://github.com/elastic/puppet-elasticsearch/pull/1111 is what's needed, no need to globally notify entire class

scornelissen85:
> I think #1111 is what's needed, no need to globally notify entire class

If you use that method then Elasticsearch will never restart after a config change, is it?

PR #1116, opened 2021-03-23
Add option to set private key type

scornelissen85:
With this PR it's now possible to set the type of the private key. If, for example, an EC private key is used, it can set with value `ec`, which will be passed to `java_ks`. The new setting is optional, when not set the current default (rsa) is used. See https://github.com/puppetlabs/puppetlabs-java_ks/blob/main/REFERENCE.md#private_key_type


phaedriel:
Integrate with https://github.com/voxpupuli/puppet-elasticsearch/pull/1186

PR #1128, opened 2021-10-27
Add support for Puppet 7

smortex:
This PR also include:

* #1126 
* #1127

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -77,7 +77,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 4.10.0 < 7.0.0"
+      "version_requirement": ">= 6.1.0 < 7.0.0"
     }
   ]
 }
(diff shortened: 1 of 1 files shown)


metadata.json
```diff
@@ -53,31 +53,26 @@
...
         "18.04",
         "20.04"
       ]
     },
     {
       "operatingsystem": "SLES",
       "operatingsystemrelease": [
-        "12.1",
-        "12.2"
       ]
```
bastelfreak, line 69: I thought SLES still has support?

smortex: According to https://www.suse.com/lifecycle 12.1 support ended on 2017-05-31 and 12.2 support ended on 2018-03-31.

Maybe we can adjust this list with `["12"]`?

bastelfreak: yes, we should switch to major versions only.

smortex: Updated in #1127

bastelfreak (approved):

PR #1140, opened 2021-12-30
fix error in custom fact resolution when conf file is empty

fe80:
Actually if a empty file exit the fact fail:
```
facter -p elasticsearch
2021-12-30 10:09:58.517175 ERROR puppetlabs.facter - error while resolving custom facts in /opt/puppetlabs/puppet/cache/lib/facter/es_facts.rb: undefined method `[]' for false:FalseClass
```

### DIFF
--- a/lib/facter/es_facts.rb
+++ b/lib/facter/es_facts.rb
@@ -50,6 +50,8 @@ module EsFacts
 
     if File.readable?("#{dir_prefix}/elasticsearch.yml")
       config_data = YAML.load_file("#{dir_prefix}/elasticsearch.yml")
+      return unless config_data
+
       httpport, ssl = get_httpport(config_data)
     end
 


smortex:
Hi!

This CI failure is relevant, can you please amend your commit?

From your working directory:

```
# Fix the reported error by adding a new line after the guard
git add lib/facter/es_facts.rb
git commit --amend     # Rewrite the last commit so that it includes the change
git push -f            # Send the changes (-f is required because we re-wrote history)
```

kenyon:
Could you also update the commit message? I think the current message is hard to parse.

fe80:
Hello,

Sorry I don't have ckeck the CI yesterday, I was doing something else.

It's ok like this ?

lib/facter/es_facts.rb
```diff
@@ -50,6 +50,8 @@ def self.run
 
     if File.readable?("#{dir_prefix}/elasticsearch.yml")
       config_data = YAML.load_file("#{dir_prefix}/elasticsearch.yml")
+      return unless config_data
```
bastelfreak, line 53: I did some testing. YAML.load_file() fails if the file doesn't exist. we already guard that case with the File.readable?. YAML.load_file returns false if the yml is broken/fails.

PR #1146, opened 2022-02-22
Add support for composable index templates

NITEMAN:
#### Pull Request (PR) description
Add support for composable index templates

PR #1148, opened 2022-03-06
Fix CI

smortex:
Older facter bundled with Puppet 6 needs lsb-release, old Ubuntu images
seems not to ship with it anymore.


smortex:
#1146 shows CI is broken on Puppet 6 on Ubuntu 18.04. I can't push a fix there so here it is.

PR #1150, opened 2022-03-15
Define default_target for parsed file providers

towo:

Due to an automated Rubocop change in efcc297b77f729ab6658201070149fb40f8805af, the module became technically unusable if using any of the types based on `ElasticParsedFile`.

By switching the `Puppet::Provider::ElasticParsedFile#xpack_config` to return/define on `self.default_target` instead of `@xpack_config`, we fix the compilation problem caused by [an exception](https://github.com/puppetlabs/puppet/blob/093c708884397a23595a2bb6ed0dfe24a261b898/lib/puppet/provider/parsedfile.rb#L381) being triggered.

n/a

### DIFF
--- a/lib/puppet/provider/elastic_parsedfile.rb
+++ b/lib/puppet/provider/elastic_parsedfile.rb
@@ -9,6 +9,6 @@ class Puppet::Provider::ElasticParsedFile < Puppet::Provider::ParsedFile
   #
   # @return String
   def self.xpack_config(val)
-    @xpack_config ||= "/etc/elasticsearch/#{val}"
+    self.default_target ||= "/etc/elasticsearch/#{val}"
   end
 end
--- a/spec/unit/provider/elastic_parsedfile_spec.rb
+++ b/spec/unit/provider/elastic_parsedfile_spec.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+
+$LOAD_PATH.unshift(File.join(File.dirname(__FILE__), '..', '..', 'lib'))
+
+require 'spec_helper_rspec'
+require 'puppet/provider/elastic_parsedfile'
+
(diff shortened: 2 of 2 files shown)


bastelfreak:
can you add a test for this? Odd that nobody complained before.

towo:
> can you add a test for this? Odd that nobody complained before.

There isn't even a test for the type at all… well.

bastelfreak (approved):

bastelfreak (approved):

smortex:
Looks like rubocop wanted to rename a variable in `lib/puppet/provider/elastic_parsedfile.rb`:

```ruby
  def self.xpack_config(val)
    @default_target ||= "/etc/elasticsearch/#{val}"
  end
```

was changed in efcc297b77f729ab6658201070149fb40f8805af to:

```ruby
  def self.xpack_config(val)
    @xpack_config ||= "/etc/elasticsearch/#{val}"
  end
```

But here we are not caching the return value of xpack_config… 

I suggest changing this code to:

```ruby
  def self.xpack_config(val)
    self.default_target ||= "/etc/elasticsearch/#{val}"
  end
```

it should not make rubocop unhappy and should fix the issue.  @towo can you please test it and confirm if it fix the issue (this is untested and I only hake legacy ES with legacy version of the module).

towo:
> it should not make rubocop unhappy and should fix the issue. @towo can you please test it and confirm if it fix the issue (this is untested and I only hake legacy ES with legacy version of the module).

Well, the patched version doesn't seem to produce any errors. Rolling back to the unpatched version makes the error reproducible, so … great success!

smortex:
**Awesome!**  Maybe we can squash the commits and reword it with the PR to better explain what we do?

Then we release a new patch version?

towo:
@smortex Here you go for review, do ping me if it needs better words™.

PR #1168, opened 2022-08-26
Add match to file_line jvm.options

phaedriel:

Add a match on file_line jvm_options (A=value) to remove old same parameter (A=oldvalue)

n/a


smortex (approved):

smortex:
file_line sucks here, so it fix some issues. I guess that if we want this to be reliable we should forget file_line and manage the whole file with a template.

PR #1177, opened 2022-12-15
Add a parameter to package mark hold

### LINKED ISSUE #1133: Package not set to status hold on Debian
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 6.17.1
- Ruby: 2.5.9p229
- Distribution: Debian 10 & 11
- Module version:  7.0.0

## How to reproduce (e.g Puppet code you use)
```
  class { 'elasticsearch':
    autoupgrade       => false,
    restart_on_change => true,
    version           => '7.15.0'
  }
```
Install elasticsearch using the module, wait for a new version of the application to be released,
run apt update && apt dist-upgrade -y
watch as elasticsearch gets upgraded to the new version even if you've specified an earlier version in
your manifest

## What are you seeing
An upgrade that may not always be wanted

## What behaviour did you expect instead
That the package was set to status hold upon installation so that a dist-upgrade
won't accidentally upgrade the package

phaedriel:

Add attribut 'mark' on resource package

Fixes #1133

### DIFF
--- a/manifests/package.pp
+++ b/manifests/package.pp
@@ -153,9 +153,19 @@ class elasticsearch::package {
   }
 
   if ($elasticsearch::package_provider == 'package') {
-    package { 'elasticsearch':
-      ensure => $package_ensure,
-      name   => $elasticsearch::_package_name,
+    # You cannot use "mark" property while "ensure" is one of ["absent", "purged", "held"]
+    if $package_ensure in ['absent', 'purged', 'held'] {
+      package { 'elasticsearch':
+        ensure => $package_ensure,
+        name   => $elasticsearch::_package_name,
+      }
+    } else {
+      # https://puppet.com/docs/puppet/7/types/package.html#package-attribute-mark
+      package { 'elasticsearch':
+        ensure => $package_ensure,
+        name   => $elasticsearch::_package_name,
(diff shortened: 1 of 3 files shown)


manifests/package.pp
```diff
@@ -153,9 +153,20 @@
...
+        ensure => $package_ensure,
+        name   => $elasticsearch::_package_name,
+      }
+    } else {
+      # https://puppet.com/docs/puppet/7/types/package.html#package-attribute-mark
+      # Set to hold to tell Debian apt/Solaris pkg to hold the package version
+      package { 'elasticsearch':
+        ensure => $package_ensure,
+        name   => $elasticsearch::_package_name,
+        mark   => ($elasticsearch::autoupgrade ? { true  => 'none', false => 'hold', }),
```
smortex, line 168: Not sure if re-using `$elasticsearch::autoupgrade` is a good idea. Of course if one wants updates to be automatically installed holding the package does not make sense, but otherwise one will probably want to know if updates are available.

What do you think about adding a new parameter like `$elasticsearch::package_hold`?

phaedriel: Ok but i will do this modification when i come back after my holidays ;) (i have only my ipad actually)

phaedriel: Done

PR #1181, opened 2022-12-29
Test binary path elasticsearch-plugin

### LINKED ISSUE #1109: Define path to plugin binary installer for older es 2 version
Hello using this module works great mostly, but unfortuently we still have an old 2.4 ES version. I noticed in the ruby.rb that the path to the plugin installer binary is hardcoded to use elasticsearch-plugin. In the change from 2 to 5 they moved this from just plugin to elasticsearch-plugin so on 2.4 we can't install plugins because it obviously can't find the binary. Is there anyway to override this or could we get a backport to allow this as an optional override to use?

File that has these values hardcoded:

https://github.com/elastic/puppet-elasticsearch/blob/master/lib/puppet/provider/elasticsearch_plugin/ruby.rb

### LINKED ISSUE #1172: elasticsearch_plugin/ruby.rb wrong path to call plugin binary
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 7.19.0
- Distribution: CentOS Linux release 7.9.2009 (Core)
- Module version: 8.0.2

## How to reproduce (e.g Puppet code you use)

Try to install a plugin 

```

   if $plugins {
     $plugin = ['delete-by-query','mapper-attachments']
     elasticsearch::plugin { $plugin: }
   }
```

This because the exec do not call the right binary to install plugin

As a workaround we linked the path in puppet code :

```
   if $plugins {
+    file { '/usr/share/elasticsearch/bin/elasticsearch-plugin':
+      ensure => link,
+      target => '/usr/share/elasticsearch/bin/plugin',
+      owner  => 'root',
+      group  => 'root',
+      mode   => '0755',
+    }
     $plugin = ['delete-by-query','mapper-attachments']
     elasticsearch::plugin { $plugin: }
   }

```
This seems to be in  elasticsearch_plugin/ruby.rb at :

```

# frozen_string_literal: true
require 'puppet/provider/elastic_plugin'
Puppet::Type.type(:elasticsearch_plugin).provide(
  :elasticsearch_plugin,
  parent: Puppet::Provider::ElasticPlugin
) do
  desc <<-END
    Post-5.x provider for Elasticsearch bin/elasticsearch-plugin
    command operations.'
  END
  case Facter.value('osfamily')
  when 'OpenBSD'
    commands plugin: '/usr/local/elasticsearch/bin/elasticsearch-plugin'
    commands es: '/usr/local/elasticsearch/bin/elasticsearch'
    commands javapathhelper: '/usr/local/bin/javaPathHelper'
  else
    commands plugin: '/usr/share/elasticsearch/bin/elasticsearch-plugin'
    commands es: '/usr/share/elasticsearch/bin/elasticsearch'
  end
end
```

Could this be fixed ? 
Thanks, best regards

phaedriel:

- Add test path binary 'plugin'

Ref : https://discuss.elastic.co/t/missing-plugin-command-in-elastic/61995/2 > "elasticsearch-plugin is only applicable for 5.0 and above. if you are really on 2.4 then you need to use /usr/share/elasticsearch/bin/plugin."

Fixes #1172
Fixes #1109

### DIFF
--- a/lib/puppet/provider/elasticsearch_plugin/ruby.rb
+++ b/lib/puppet/provider/elasticsearch_plugin/ruby.rb
@@ -17,7 +17,11 @@ Puppet::Type.type(:elasticsearch_plugin).provide(
     commands es: '/usr/local/elasticsearch/bin/elasticsearch'
     commands javapathhelper: '/usr/local/bin/javaPathHelper'
   else
-    commands plugin: '/usr/share/elasticsearch/bin/elasticsearch-plugin'
+    if File.exist? '/usr/share/elasticsearch/bin/elasticsearch-plugin'
+      commands plugin: '/usr/share/elasticsearch/bin/elasticsearch-plugin'
+    else
+      commands plugin: '/usr/share/elasticsearch/bin/plugin'
+    end
     commands es: '/usr/share/elasticsearch/bin/elasticsearch'
   end
 end


smortex:
Do you really want to fix/add support for ES < 5? As far as I can tell, anything older than 7.13.x is not supported:

https://www.elastic.co/fr/support/eol

Maybe we can just close the issue?

phaedriel:
> Do you really want to fix/add support for ES < 5? As far as I can tell, anything older than 7.13.x is not supported:
> 
> https://www.elastic.co/fr/support/eol
> 
> Maybe we can just close the issue?

- solution 1 : we add this small modification (i could add a comment to better explain in ruby code)
- solution 2 : i rollback and we suppress this indication "2.x" in readme (module description) "This module is actively tested against Elasticsearch 2.x, 5.x, and 6.x."

as you prefer ;)

bastelfreak (approved):
I think this is fine. This doesn't hardcode any paths with a version in it. It's possible that distros buil their own elasticsearch packages and use the old path.

smortex (approved):

PR #1071, opened 2020-02-19
update log4j2properties.erb for [%node_name]%markers 

sferry:
Pull request acceptance prerequisites:

- [ ] Signed the [CLA](https://www.elastic.co/contributor-agreement/) (if not already signed)
- [ ] Rebased/up-to-date with base branch
- [ ] Tests pass
- [ ] Updated CHANGELOG.md with patch notes (if necessary)
- [ ] Updated CONTRIBUTORS (if attribution is requested)

PR #1185, opened 2022-12-31
Add support for log4j2.properties and unused parameters become deprecated (daily_rolling_date_pattern , default_logging_level, file_rolling_type, logging_file, logging_level, logging_template, deprecation_logging, deprecation_logging_level, rolling_file_max_backup_index, rolling_file_max_file_size, security_logging_content, security_logging_source)

### LINKED ISSUE #529: log4j config, how to add custom appender
We are looking to add a modification to the log4j appender type used to get the rolling file appender vs the daily rolling file appender.

http://grokbase.com/t/gg/elasticsearch/14chrn08rj/pruning-logs-how-to-get-maxbackupindex-to-work

From what I see in the [template](https://github.com/elastic/puppet-elasticsearch/blob/master/templates/etc/elasticsearch/logging.yml.erb#L13-L15) the only change we can make is to alter the already-configured appender. Is that right?

Is there a way to inject a modified appender via hiera? I'm not sure if this is the right yml structure for hieradata but this is what I came up with to try out/test:

```
    logging_config:
      rollingFile:
        file:
          type: rollingFile
          file: "${path.logs}/${cluster.name}.log"
          maxBackupIndex: 10
          layout:
            type: pattern
            conversionPattern: '"[%d{ISO8601}][%-5p][%-25c] %m%n"'
```

Not sure if the nested yml structure is the right way to go. The elasticsearch.yml renders something like `file.type: rollingFile` instead

### LINKED ISSUE #907: Same template for both logging.yml and log4j2.properties
Hi,

https://github.com/elastic/puppet-elasticsearch/blob/6d31554205305d4127acb8d31f4e640466c42f70/manifests/instance.pp#L244
https://github.com/elastic/puppet-elasticsearch/blob/6d31554205305d4127acb8d31f4e640466c42f70/manifests/instance.pp#L245

Is there any reason that only one template is allowed for both logging.yml and log4j2.properties ? When using default values, there are 2 templates (which is correct), but if you override the logging_template variable, it will replace both files, while one is YML, the other is ini like, and both have totally different content.

### LINKED ISSUE #1056: ESJsonLayout as layout type
How to set `appender.rolling.layout.type` parameter to `ESJsonLayout` in log4j.properties?

phaedriel:

- Add indication deprecated parameters (if not used) :  daily_rolling_date_pattern, default_logging_level , file_rolling_type, logging_file, logging_level, logging_template, deprecation_logging, deprecation_logging_level, rolling_file_max_backup_index, rolling_file_max_file_size, security_logging_content, security_logging_source
- Suppress template log4j2.properties.erb
- Modify manifest config.pp
- Add function ruby es_hash2properties

- Fixes #529
- Fixes #907 
- Fixes #1056 

Replace Opened PR

- #1021
- #1071


smortex:
@phaedriel thanks for this PR!

Just to be sure, this should be tagged as a breaking change? Can you reword the PR title so that it "speaks" to end-user when they read the ChangeLog (which use the PR title and link to it for further details)?

Thanks!

smortex:
> Better ?

Thanks!

> I have keep unused parameters, just move them on bottom list and add optional if necessary > remove them will be more easy in future release. I have clean unused values in data yaml files.

I saw, and its awesome :heart_eyes: 

> I'm not sure it's a breaking change ?

I used an old version of the module (ES 6.x, when the module was managing multiple instances) and it seems it managed log4j2.properties. But looking at the changes, it seems it stopped doing so, so that's why I am asking.

If I am right, this in an "enhancement" (it add a new feature: previously in this major branch the file could not be managed and now it can be). But if I update from a previous version (same major version, let's say the latest version 8.0.2), if it will change the content of the file it is a breaking change (I don't think we are in this case). Maybe the user could have some settings in `logging_config` that where ignored and will be applied again, this is not a breaking change, at most it is an enhancement, and someone can argue it is a bugfix :-)

PR #1186, opened 2022-12-31
Add option to set private key type

phaedriel:

- Finish PR #1116
"With this PR it's now possible to set the type of the private key. If, for example, an EC private key is used, it can set with value ec, which will be passed to java_ks. The new setting is optional, when not set the current default (rsa) is used. See https://github.com/puppetlabs/puppetlabs-java_ks/blob/main/REFERENCE.md#private_key_type"

No one

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -205,6 +205,9 @@
 # @param private_key
 #   Path to the key associated with this node's certificate.
 #
+# @param private_key_type
+#   The type of the private key. Usually the private key is of type RSA key but it can also be an Elliptic Curve key (EC) or DSA.
+#
 # @param proxy_url
 #   For http and https downloads, you may set a proxy server to use. By default,
 #   no proxy is used.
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -205,6 +205,10 @@
 # @param private_key
 #   Path to the key associated with this node's certificate.
 #
+# @param private_key_type
+#   The type of the private key. Usually the private key is of type RSA key but it can also be an Elliptic Curve key (EC) or DSA.
+#   Valid values: rsa, dsa, ec
```
kenyon, line 210: Data type will be shown in the generated REFERENCE.md, so no need to repeat it here.
```suggestion
```

kenyon (changes requested):

manifests/init.pp
phaedriel: ok, i suppress this line and i rebase my branch

kenyon (approved):

smortex (approved):

PR #1192, opened 2023-05-12
modulesync 7.3.0; allow stdlib 9.x

bastelfreak:
modulesync 6.0.0


metadata.json
```diff
@@ -22,7 +22,7 @@
     },
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.13.0 < 9.0.0"
+      "version_requirement": ">= 4.13.0 < 10.0.0"
```
h-haaks, line 25: Had to allow 9.x cause without it `puppet module install` in acceptance never finishes ...

zilchms:
@bastelfreak should we add the enhancement label, too, since we add stdlib9 support?

PR #1194, opened 2023-06-23
only create package dir when needed

bugfood:

Only create package dir when needed.
If packages are not to be downloaded, then the package dir is unnecessary cruft.


marcusdots:
I do not have enough background of Elasticsearch to understand the impact of your change. If there was an acceptance test I would rely on it.
The benefit of not creating a directory does not make learning Elasticsearch appealing to me.

Issue #928, opened 2018-05-02
The module deletes main jvm.options file.


alexizmailov:
* Module version: 6.2.2
* Puppet version: v5.3.2
* OS and version: Centos 7.4
* ES version: v6.2.4

## Bug description

elasticsearch/manifests/config.pp:66 removes **/etc/elasticsearch/jvm.options**
```
'/etc/elasticsearch/jvm.options':
    ensure => 'absent';
```
ES can't start after that:
```
elasticsearch[53952]: Exception in thread "main" java.nio.file.NoSuchFileException: /etc/elasticsearch/jvm.options
.....
systemd[1]: elasticsearch-int-cluster.service: main process exited, code=exited, status=1/FAILURE
```
The file can be empty but it definitely should stay there.

tylerjl:
I ran a test manifest to attempt to reproduce your situation:

```
[root@centos-7-x64 ~]# cat test.pp
class { 'elasticsearch': version => '6.2.4' }
elasticsearch::instance { 'es-01' : }
[root@centos-7-x64 ~]# curl localhost:9200
{
  "name" : "centos-7-x64-es-01",
  "cluster_name" : "elasticsearch",
  "cluster_uuid" : "o6lDFbNKQ92U8fQehhHXJg",
  "version" : {
    "number" : "6.2.4",
    "build_hash" : "ccec39f",
    "build_date" : "2018-04-12T20:37:28.497551Z",
    "build_snapshot" : false,
    "lucene_version" : "7.2.1",
    "minimum_wire_compatibility_version" : "5.6.0",
    "minimum_index_compatibility_version" : "5.0.0"
  },
  "tagline" : "You Know, for Search"
}
[root@centos-7-x64 ~]# systemctl is-active elasticsearch-es-01
active
[root@centos-7-x64 ~]#
```

Things appear to work fine. Do you have any more information regarding your failure? As far as I can see, things work as expected.

alexizmailov:
I started collecting the data you requested and the problem suddenly disappeared after next puppet run.

I rebuilt the node and created another fresh one but the problem has not reoccured. 

I followed all steps I did yesterday and it just works today. The only difference is my new manager - he was watching me from behind the shoulder yesterday, so it might be his bad aura.

Sorry for bothering.

tylerjl:
No worries, glad it's working!

KoenDierckx:
I've also run into this issue. Any thoughts how to resolve this?
Accept #1041 ?

PR #1198, opened 2023-07-11
fixes #1197 - add a managed file entry for jvm.options

### LINKED ISSUE #1093: 'jvm.options' (prerequisite, required file) is unmanaged by module
## Bug description
The file "jvm.options" is not managed by the current 'master' branch of the module, however it's required for an Augueas lens in manifests/config.pp lines 185 - 193.

When installing new nodes, this file is present and the current Augueas lens works as desired.

When migrating clients from the legacy Forge version of the module (v6.4.0) this file is deliberately removed by the module (config.pp, lines 72,73), and as such never put into place again to make everything work correctly, unless done manually.

In addition to being able to override the default provided array of options from 'common.yaml', the Java Memory parameters should be available to configure as top level parameters.

### LINKED ISSUE #1197: File jvm.options in root is unmanaged/may be missing
This effectively re-opens #928 and relates to #1041.

When migrating from the multi instance to single instance of elasticsearch, the `jvm.options` file may not exist and is not created by the module.  This results in elasticsearch being unable to start.  As pointed out in #928 the file may be empty but does need to exist.  The manual work around is to `touch /etc/elasticsearch/jvm.options` so the file does exist and allows the service to start.  I am willing to guess the file is indeed present in the default install and that is why this doesn't trigger an issue more often.

from `/var/log/messages`
```
Jul  9 12:22:44 es-0001 systemd: Starting Elasticsearch...
Jul  9 12:22:45 es-0001 systemd-entrypoint: Exception in thread "main" java.nio.file.NoSuchFileException: /etc/elasticsearch/jvm.options
Jul  9 12:22:45 es-0001 systemd-entrypoint: at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
Jul  9 12:22:45 es-0001 systemd-entrypoint: at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
Jul  9 12:22:45 es-0001 systemd-entrypoint: at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
Jul  9 12:22:45 es-0001 systemd-entrypoint: at java.base/sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:218)
Jul  9 12:22:45 es-0001 systemd-entrypoint: at java.base/java.nio.file.Files.newByteChannel(Files.java:380)
Jul  9 12:22:45 es-0001 systemd-entrypoint: at java.base/java.nio.file.Files.newByteChannel(Files.java:432)
Jul  9 12:22:45 es-0001 systemd-entrypoint: at java.base/java.nio.file.spi.FileSystemProvider.newInputStream(FileSystemProvider.java:422)
Jul  9 12:22:45 es-0001 systemd-entrypoint: at java.base/java.nio.file.Files.newInputStream(Files.java:160)
Jul  9 12:22:45 es-0001 systemd-entrypoint: at org.elasticsearch.tools.launchers.JvmOptionsParser.readJvmOptionsFiles(JvmOptionsParser.java:168)
Jul  9 12:22:45 es-0001 systemd-entrypoint: at org.elasticsearch.tools.launchers.JvmOptionsParser.jvmOptions(JvmOptionsParser.java:124)
Jul  9 12:22:45 es-0001 systemd-entrypoint: at org.elasticsearch.tools.launchers.JvmOptionsParser.main(JvmOptionsParser.java:86)
Jul  9 12:22:46 es-0001 systemd: elasticsearch.service: main process exited, code=exited, status=1/FAILURE
Jul  9 12:22:46 es-0001 systemd: Failed to start Elasticsearch.
Jul  9 12:22:46 es-0001 systemd: Unit elasticsearch.service entered failed state.
Jul  9 12:22:46 es-0001 systemd: elasticsearch.service failed.
```

bmagistro:
This fixes #1197 and effectively replaces #1041 which was originally opened to support #928.  I believe this will also fix #1093.

The jvm.options file is considered a config by the RPM (would expect the same to be true of a DEB) so when migrating from a multi-instance to a single instance management and not changing versions, its possible the file `jvm.options` would not exist in the new configuration directory.  This change ensures the file exists and if not creates an empty file.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -181,6 +181,15 @@ class elasticsearch::config {
       mode    => '0440',
     }
 
+    file { "${elasticsearch::configdir}/jvm.options":
+      ensure  => 'file',
+      notify  => $elasticsearch::_notify_service,
+      require => Class['elasticsearch::package'],
+      owner   => $elasticsearch::elasticsearch_user,
+      group   => $elasticsearch::elasticsearch_group,
+      mode    => '0640',
+    }
+
     if ($elasticsearch::version != false and versioncmp($elasticsearch::version, '7.7.0') >= 0) {
       # https://www.elastic.co/guide/en/elasticsearch/reference/master/advanced-configuration.html#set-jvm-options
       # https://github.com/elastic/elasticsearch/pull/51882


h-haaks:
Fixed by #1202

Issue #1153, opened 2022-03-28
elasticsearch_keystore resource fails due to legacy instance code


johnkazmerzak:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 6.26.0
- Ruby: 2.5.9p229
- Distribution: CentOS 7
- Module version: 8.0.1

## How to reproduce (e.g Puppet code you use)

```
elasticsearch_keystore { $hostname:
  settings => {
    'bootstrap.password' => $bootstrap_password,
    'xpack.security.transport.ssl.keystore.secure_password' => $keystore_truststore_secure_password,
    'xpack.security.transport.ssl.truststore.secure_password' => $keystore_truststore_secure_password,
  },
}
```

## What are you seeing
Could not evaluate: Execution of '/usr/share/elasticsearch/bin/elasticsearch-keystore create' returned 1:

## What behaviour did you expect instead
Successful execution of elasticsearch_keystore resource

## Output log
```
Notice: /Stage[main]/Sstsrdnalin_profiles::Elastic/Elasticsearch_keystore[redacted]/ensure: created
Debug: Executing with uid=elasticsearch gid=elasticsearch: '/usr/share/elasticsearch/bin/elasticsearch-keystore create'
Error: /Stage[main]/Sstsrdnalin_profiles::Elastic/Elasticsearch_keystore[dnatestlog03]: Could not evaluate: Execution of '/usr/share/elasticsearch/bin/elasticsearch-keystore create' returned 1: 
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/util/execution.rb:297:in `execute'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/provider.rb:106:in `execute'
/opt/puppetlabs/puppet/cache/lib/puppet/provider/elasticsearch_keystore/ruby.rb:49:in `run_keystore'
/opt/puppetlabs/puppet/cache/lib/puppet/provider/elasticsearch_keystore/ruby.rb:98:in `flush'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/type.rb:1045:in `flush'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/transaction/resource_harness.rb:25:in `evaluate'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/transaction.rb:267:in `apply'
[...]

Issue #1108, opened 2021-01-20
V7.0.0 of this module struggling with keystore creation


nick-george:
I'm wondering if this is the same issue as #983. 

Note, I'm using ES (OSS RPM version on Oracle Linux 7.9)

Here's my call into the ES module: 

```
  $es_secrets = {
      'keystore.seed' => '<redacted>',
  }

  class { '::elasticsearch':
    ensure            => 'present',
    version           => $version,
    autoupgrade       => false ,
    manage_repo       => false,
    require           => [Package['java',],File_line['block-es-package-obsoletion']],
    status            => 'enabled',
    restart_on_change => true,
    config            => $config,
    api_timeout       => 30, #increase the timeout for a slow starting ES instance.
    package_name      => $package_name,
    jvm_options       => [
      "-Xms${heapstr}",
      "-Xmx${heapstr}"
    ],
    secrets           => $es_secrets,
    datadir           => '/data/elasticsearch',
    #keystore_password => '<password_redacted>', #passphrase only appears to be used by xpack?
  }
```
### Symptoms:
* If I delete the keystore, the module runs without issue. 
* If I let the module create the keystore, then run puppet again, I get the following error

``` 
Debug: Prefetching elasticsearch_keystore resources for elasticsearch_keystore
Notice: /Stage[main]/Elasticsearch::Config/Elasticsearch_keystore[elasticsearch_secrets]/ensure: created (corrective)
Debug: Executing with uid=elasticsearch gid=elasticsearch: '/usr/share/elasticsearch/bin/elasticsearch-keystore create'
Error: /Stage[main]/Elasticsearch::Config/Elasticsearch_keystore[elasticsearch_secrets]: Could not evaluate: Execution of '/usr/share/elasticsearch/bin/elasticsearch-keystore create' returned 1: 
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/util/execution.rb:297:in `execute'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/provider.rb:106:in `execute'
/opt/puppetlabs/puppet/cache/lib/puppet/provider/elasticsearch_keystore/ruby.rb:47:in `run_keystore'
/opt/puppetlabs/puppet/cache/lib/puppet/provider/elasticsearch_keystore/ruby.rb:96:in `flush'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/type.rb:1045:in `flush'
[...]

Issue #983, opened 2018-09-13
elasticsearch.keystore generated at every run


ml14tlc:
* 6.3.3:
* 4.10 and 5.5:
* RedHat 7.5 and CentOS 7.3:

## Bug description

I'm using the _secrets_ parameter to manage the file _elasticsearch.keystore_. Every time Puppet runs the file is recreated, even when the parameter _purge\_secrets_ is set to _true_. In particular, Puppet wants to remove _keystore.seed_. I suppose a new keystore is created each time, which implies the presence of _keystore.seed_, the entries in the config file are added and the temporary file is compared with the current one.

Besides that, would it be possible to 'unmanage' _elasticsearch.keystore_? Even if _secrets_ is left to _undefined_ and _elasticsearch.keystore_, Puppet replaces it with an "empty" one (except for the _keystore.seed_ value). I'd prefer generate the file locally and the distribute it, rather than having passwords in the manifest.

tylerjl:
Hmm, that behavior is definitely unexpected, the keystore provider should be intelligent enough to a) not try to create keystores and b) only detect values that you _want_ in the keystore that it should change and leave others alone.

If you want to dig into the unexpected behavior a bit more to try and find the bug, I'd love to see an example manifest that demonstrates the behavior.

With regards to leaving the keystore unmanaged entirely, the module only manages the `elasticsearch_keystore` resource if you've set secrets (as you mention, if `secrets` is defined), so the file is coming back, it'd have to be happening from the `elasticsearch` executable or some other method since the module will only try to manage it if `$elasticsearch::secrets` or `$secrets` on the `elasticsearch::instance` defined type is set.

tylerjl:
@ml14tlc thank you for the configuration examples, that's super helpful to look at the problem.

The configuration files coupled with your logs seem to make sense to me. Elasticsearch sets a seed value for the keystore, and because your configuration is set to purge secrets, since `seed` isn't present in your instance config's `secrets` parameter, the module tries to remove it.

Out of curiosity, does the module stop mutating the keystore if you change `purge_secrets` from `true` to `false`? I may just be misunderstanding something, since your latest comment suggests that you don't want the keystore to be created at all, but your configuration file is setting secrets (that require a keystore).

ml14tlc:
Hi @tylerjl 

* if I remove the option _purge\_secrets_ to false the keystore is still re-created at every run, of course the message is a bit different, but consistent (_would have removed ..._):
```
[...]

PR #1199, opened 2023-07-12
Optional keystore mgmt

bmagistro:
This change makes managing the secrets keystore optional.  Since the code in that is still setup for instance based deployments which are no longer supported it causes failures when utilizing this module.  By making the secrets keystore management optional it buys time to fix that code while still allowing users to take advantage of the other managed aspects.

This PR is based off of #1198 as we needed both changes in our environment.

#### This Pull Request (PR) likely relates to the following issues
#1153 
#1108 
#983

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -229,7 +229,7 @@ class elasticsearch::config {
     }
 
     # Add secrets to keystore
-    if $elasticsearch::secrets != undef {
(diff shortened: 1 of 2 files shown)


marcusdots:
I do not have enough context of Elasticsearch to understand the impact of `jvm.options`. If it had an acceptance test I would simply rely on it.

PR #1208, opened 2023-12-04
stdlib v9 support, use namespaced stdlib functions

ggabijaa:
#### Pull Request (PR) description
Added Stdlib v9 support
Added Puppet v8 support

#### This Pull Request (PR) fixes the following issues
Fixes https://github.com/voxpupuli/puppet-elasticsearch/issues/1201

PR #1213, opened 2024-02-21
Drop Ubuntu 18.04 support

zilchms:
includes: #1212


h-haaks:
Drop Debian 10 as well?

h-haaks:
I am planning to do a minor release with stdlib 9.x support now, and then fix and merge #1208 which is also a breaking change.
Ok to wait merging this until then?

PR #1215, opened 2024-02-21
Release 9.0.0

h-haaks:


bastelfreak:
It would be nice if we could include https://github.com/voxpupuli/puppet-elasticsearch/pull/1208 as well

bastelfreak:
As mentioned in the other PR, it can be reworked to use native puppet code, so it doesn't depend on stdlib.

bastelfreak (approved):

Issue #1201, opened 2023-08-07
warnings when used with stlib 9.


robertc99:
Module isnt fully compatible with stdlib 9.
Generates the following warning

Warning: This function is deprecated, please use stdlib::merge instead. at ["/etc/puppetlabs/code/environments/elk_staging/r10k/modules/elasticsearch/manifests/config.pp", 24]:["/etc/puppetlabs/code/environments/elk_staging/r10k/modules/elasticsearch/manifests/init.pp", 476]
   (location: /etc/puppetlabs/code/environments/elk_staging/r10k/modules/stdlib/lib/puppet/functions/deprecation.rb:35:in `deprecation')

h-haaks:
fixed by #1217

PR #1217, opened 2024-02-21
Remove stdlib dependency

h-haaks:
This module only used stdlib:merge which is now replaced with '+'

#1201

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -21,7 +21,7 @@ class elasticsearch::config {
 
   $init_defaults = {
     'MAX_OPEN_FILES' => '65535',
-  }.merge($elasticsearch::init_defaults)
+  } + $elasticsearch::init_defaults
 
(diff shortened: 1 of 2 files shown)


zilchms (approved):

manifests/config.pp
```diff
@@ -164,12 +164,8 @@
...
 
     # Generate Elasticsearch config
-    $data = merge(
-      $elasticsearch::config,
-      { 'path.data' => $elasticsearch::datadir },
-      { 'path.logs' => $elasticsearch::logdir },
-      $_tls_config
-    )
+    $data =
+      $elasticsearch::config + { 'path.data' => $elasticsearch::datadir } + { 'path.logs' => $elasticsearch::logdir } + $_tls_config
```
zilchms, line 168: i think you could put all of this into one line, but at that point its just nitpicking  :)

PR #1224, opened 2024-05-04
Modify array setting[6..] to setting[6..setting.length - 1] in ruby code

phaedriel:

- Replace in ruby files array setting[6..] by setting[6..setting.length - 1]
- Replace a ':' by '=' in self.rest method (elastic_rest.rb)

Issue https://github.com/voxpupuli/puppet-elasticsearch/issues/1222

### DIFF
--- a/lib/puppet/type/elasticsearch_component_template.rb
+++ b/lib/puppet/type/elasticsearch_component_template.rb
@@ -48,7 +48,7 @@ Puppet::Type.newtype(:elasticsearch_component_template) do
               val['template']['settings']['index'] = {} unless val['template']['settings'].key? 'index'
               (val['template']['settings'].keys - ['index']).each do |setting|
                 new_key = if setting.start_with? 'index.'
-                            setting[6..]
+                            setting[6..setting.length - 1]
                           else
                             setting
                           end
--- a/lib/puppet/type/elasticsearch_index_template.rb
+++ b/lib/puppet/type/elasticsearch_index_template.rb
@@ -54,7 +54,7 @@ Puppet::Type.newtype(:elasticsearch_index_template) do
                 val['template']['settings']['index'] = {} unless val['template']['settings'].key? 'index'
                 (val['template']['settings'].keys - ['index']).each do |setting|
                   new_key = if setting.start_with? 'index.'
-                              setting[6..]
+                              setting[6..setting.length - 1]
                             else
                               setting
                             end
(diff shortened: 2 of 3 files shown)


lib/puppet/provider/elastic_rest.rb
```diff
@@ -33,7 +33,7 @@ def self.rest(http,
                 timeout = 10,
                 username = nil,
                 password = nil,
-                validate_tls: true)
+                validate_tls = true)
```
smortex, line 36: This change the semantic and is probably not what you want (it seems to not break tests so this demonstrate that unit tests are missing).

You can read more about "keyword arguments" here if some context is needed:
https://www.rubyguides.com/2018/06/rubys-method-arguments/

For now, I would suggest adding a unit test that demonstrate that the current behavior is wrong in a first commit, and then fix it with this change (2 commits in the PR).

lib/puppet/type/elasticsearch_component_template.rb
```diff
@@ -48,7 +48,7 @@ def insync?(value)
               val['template']['settings']['index'] = {} unless val['template']['settings'].key? 'index'
               (val['template']['settings'].keys - ['index']).each do |setting|
                 new_key = if setting.start_with? 'index.'
-                            setting[6..]
+                            setting[6..setting.length - 1]
```
smortex, line 51: This is a longer way to say:
```ruby
setting[6..-1]
```

which is a longer way to say (original code):
```ruby
setting[6..]
```

What was intended here?

smortex (changes requested):

lib/puppet/type/elasticsearch_component_template.rb
phaedriel: - [6..-1] generate Rubo Cop:: Cop:: Style:: Slicing With Range error
- [6..] generate error issue https://github.com/voxpupuli/puppet-elasticsearch/issues/1222

I try to find a solution

lib/puppet/provider/elastic_rest.rb
phaedriel: i rollback

binford2k:
Hallo @phaedriel! Fill out this form if you'd like to claim your May the Source Be With You sticker! https://forms.office.com/r/Cn55uJmWMH

lib/puppet/type/elasticsearch_component_template.rb
smortex: I see. This was caused by EOL Ruby bundled with EOL Puppet, better stick to "modern standards" rather than trying to be compatible with EOL products.  Older versions can still be used if necessary.

PR #1235, opened 2024-09-10
Set xpack config to enable authentication using username and password

miluxhd:
This pull request adds a `password_enabled` parameter to the Elasticsearch Puppet module to enable user/password authentication.

as there's no `xpack.security.enabled` configuration in this module I can not enable username/password authentication for Elasticsearch. So I added a parameter to make it configurable.


h-haaks:
There is just to many configuration options to have all of them as params in init.pp.
Thats why the config param is there.

